PR #11, opened 2021-11-05
Pia 1948 bank sdk

zladzeyka:


a-szotyori (approved):
Nice! I found a few to-dos, but they can be done in other PRs:
* Fix `Bank` example app.
* Clean up targets:
  * Merge `GiniBankSDKExampleTests` with `Example_Tests`.
  * Rename targets to use the same pattern as our other example apps: for example rename `Example Swift` to `GiniBankSDKExampleSwift` or `Bank` to `GiniBankSDKExampleBank`. 
  * Fix `Supported formats` screen: 
    * Section header is under the navigation bar and is only partially visible.
    * Icons are not positioned correctly. They are in the top left of their cell and are cut off. Tested on a device with iPhone 11 iOS 14.6 and on a simulator with iOS 15.

PR #14, opened 2021-11-10
Pia 1981 example apps

zladzeyka:
* Rename bank example project, resource folder, and targets
  * Rename health example project, resource folder, and targets
  * Update check workflows for BankSDK and HealthSDK
  * Remove # if pinning from HealthAPILibrary
PS Sorry for such a big PR


a-szotyori (approved):
Nicely done!

I pushed two commits to fix two small issues in the Health SDK's example app related to `Start with test document`.

PR #20, opened 2021-11-25
Pia 1817 bank selection button

zladzeyka:
- Add edit icon.
- Update GiniHealthConfiguration.
- Add tap action with animation on the bank selection view.
- Changed layout: bank selection and "weiter" buttons are now with the same size.


a-szotyori (approved):
Nice!

PR #39, opened 2022-02-01
fix(GiniBankSDK): Fix Return Assistant storyboard and line item xib crashes

a-szotyori:
Since iOS 15 and Xcode 13 a crash started happening in UITableViewCells related to the autoresizing
masks. By removing `translatesAutoresizingMaskIntoConstraints="NO"` from the line item view cell
xib's (`DigitalLineItemTableViewCell.xib`) root view and shadow view the crash was fixed.
Solution source: https://stackoverflow.com/a/69439370/276129

Due to switching to Swift Packages some referenced classes in the storyboard and xib had to be
updated. Also getting the bundle instance had to be updated.


zladzeyka (approved):
Thank you for fixing it! 👏🏻

PR #47, opened 2022-02-16
Pia 2267

zladzeyka:
feat(GiniBankSDK): Add amountParsingError and check if the amount is convertible to a `Double`.

Currently, we use only `EUR` as a currency.
The amount string in the `PaymentInfo` must be convertible to a `Double`.
For ex. "12.39" is valid, but "12.39 €" or "12,39" are not valid.

### DIFF
--- a/BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBank.swift
+++ b/BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBank.swift
@@ -68,7 +68,7 @@ import GiniCaptureSDK
         do {
             amountString = try String.parseAmountStringToBackendFormat(string: paymentInfo.amount)
         } catch {
-            completion(.failure(.amountParsingError))
+            completion(.failure(.amountParsingError(amountString: paymentInfo.amount)))
         }
         paymentService.resolvePaymentRequest(id: paymentRequesId, recipient: paymentInfo.recipient, iban: paymentInfo.iban, amount: amountString, purpose: paymentInfo.purpose, completion: { result in
             DispatchQueue.main.async {
--- a/BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBank.swift
+++ b/BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBank.swift
@@ -53,17 +53,24 @@ import GiniCaptureSDK
 
     /**
      Resolves the payment via payment request id.
-
+     **Important**: The amount string in the `PaymentInfo` must be convertible to a `Double`.
+           * For ex. "12.39" is valid, but "12.39 €" or "12,39" are not valid.
      - parameter paymentRequesId: Id of payment request.
      - parameter completion: An action for processing asynchronous data received from the service with Result type as a paramater. Result is a value that represents either a success or a failure, including an associated value in each case.
      Completion block called on main thread.
      In success returns the resolved payment request structure.
      In case of failure error from the server side.
+     If the amount string could not be parsed throws `amountParsingError`
 
      */
     public func resolvePaymentRequest(paymentRequesId: String, paymentInfo: PaymentInfo, completion: @escaping (Result<ResolvedPaymentRequest, GiniBankError>) -> Void) {
-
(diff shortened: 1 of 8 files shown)


BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBank.swift
```diff
@@ -53,17 +53,24 @@ import GiniCaptureSDK
...
     public func resolvePaymentRequest(paymentRequesId: String, paymentInfo: PaymentInfo, completion: @escaping (Result<ResolvedPaymentRequest, GiniBankError>) -> Void) {
-
-        paymentService.resolvePaymentRequest(id: paymentRequesId, recipient: paymentInfo.recipient, iban: paymentInfo.iban, amount: paymentInfo.amount, purpose: paymentInfo.purpose, completion: { result in
+        var amountString = ""
+        do {
+            amountString = try String.parseAmountStringToBackendFormat(string: paymentInfo.amount)
+        } catch {
+            completion(.failure(.amountParsingError(amountString: paymentInfo.amount)))
+        }
+        paymentService.resolvePaymentRequest(id: paymentRequesId, recipient: paymentInfo.recipient, iban: paymentInfo.iban, amount: amountString, purpose: paymentInfo.purpose, completion: { result in
```
a-szotyori, line 73: This line is reached even if the exception was caught and completion was called with failure.

What do you think of this alternative? It forwards the GiniBankError in the completion and warns us if we throw an unexpected error ([assertionFailure()](https://www.swiftbysundell.com/articles/picking-the-right-way-of-failing-in-swift/#lets-start-with-a-list) is ignored in release builds and only affects debug builds):
```
        var amountString: String?
        do {
            amountString = try String.parseAmountStringToBackendFormat(string: paymentInfo.amount)
        } catch let error as GiniBankError {
            completion(.failure(error))
        } catch {
            assertionFailure("Unexpected error: \(error)")
        }
        
        guard let amountString = amountString else {
            return
        }
        
        paymentService.resolvePaymentRequest(id: paymentRequesId, recipient: paymentInfo.recipient, iban: paymentInfo.iban, amount: amountString, purpose: paymentInfo.purpose, completion: { result in
            DispatchQueue.main.async {
                switch result {
                case let .success(resolvedPayment):
                    completion(.success(resolvedPayment))
                case let .failure(error):
                    completion(.failure(.apiError(error)))
                }
            }
        })
```

BankSDK/GiniBankSDK/Tests/GiniBankSDKTests/GiniBankSDKTests.swift
```diff
@@ -10,6 +10,44 @@ import XCTest
 @testable import GiniBankSDK
 
 final class GiniBankSDKTests: XCTestCase {
```
a-szotyori, line 12: The test class name is too generic. You could rename it to `String+UtilsTests` for example.

BankSDK/GiniBankSDK/Tests/GiniBankSDKTests/GiniBankSDKTests.swift
```diff
@@ -10,6 +10,44 @@ import XCTest
...
+    
+    func testParseAmountStringToBackendFormat5(){
+        let amountToPay = "28.10:EUR"
+        XCTAssertNotEqual(try? String.parseAmountStringToBackendFormat(string: amountToPay), "28.1:EUR")
+    }
+    
+    func testParseAmountStringToBackendFormat6(){
+        let amountToPay = "28,10"
+        XCTAssertNotEqual(try? String.parseAmountStringToBackendFormat(string: amountToPay), "28.1:EUR")
     }
```
a-szotyori, line 52: A test verifying that the right error is thrown is missing. The error throwing is part of the `parseAmountStringToBackendFormat()` method's public api and it's best to consolidate this behaviour with a test.

a-szotyori (changes requested):
Nicely done and thanks for adding tests! I found a bug and suggested improving the tests a bit.

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBank.swift
zladzeyka: @a-szotyori Looks good, would you like to commit this suggestion?

zladzeyka: I'll do a change

a-szotyori (approved):

PR #48, opened 2022-02-17
Pia 2214

zladzeyka:
Tracking:
 *  Add payment provider name to the  BankSelectionButtonClicked tracking event.

### DIFF
--- a/HealthSDK/GiniHealthSDK/Sources/GiniHealthSDK/Core/PaymentReviewViewController.swift
+++ b/HealthSDK/GiniHealthSDK/Sources/GiniHealthSDK/Core/PaymentReviewViewController.swift
@@ -111,8 +111,7 @@ public final class PaymentReviewViewController: UIViewController, UIGestureRecog
                 if let paymentProviders = self?.paymentProviders, paymentProviders.count > 0 {
                     let providerId = UserDefaults.standard.string(forKey: "ginihealth.defaultPaymentProviderId")
                     let provider = paymentProviders.first(where: { $0.id == providerId }) ?? paymentProviders[0]
-                    self?.configureBankProviderView(paymentProvider: provider)
-                    self?.configurePayButton(paymentProvider: provider)
+                    self?.selectedPaymentProvider = provider
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


HealthSDK/GiniHealthSDKExample/GiniHealthSDKExample/AppCoordinator.swift
```diff
@@ -67,6 +67,7 @@ final class AppCoordinator: Coordinator {
     }
     
     func start() {
+        try! apiLib.removeStoredCredentials()
```
a-szotyori, line 70: Did you commit this by accident? If we keep this in the example, then clients might think they should always call this. This would cause new users to be generated every time.

This method should be called if the client id or base url changed or there is some issue regarding authentication.

HealthSDK/GiniHealthSDK/Sources/GiniHealthSDK/Core/PaymentReviewViewController.swift
```diff
@@ -111,8 +111,7 @@ public final class PaymentReviewViewController: UIViewController, UIGestureRecog
                 if let paymentProviders = self?.paymentProviders, paymentProviders.count > 0 {
                     let providerId = UserDefaults.standard.string(forKey: "ginihealth.defaultPaymentProviderId")
                     let provider = paymentProviders.first(where: { $0.id == providerId }) ?? paymentProviders[0]
-                    self?.configureBankProviderView(paymentProvider: provider)
-                    self?.configurePayButton(paymentProvider: provider)
```
a-szotyori, line 115: These two removed method calls look important. Are you sure they are not needed anymore?

a-szotyori (changes requested):
Looks great. I have only two observations.

HealthSDK/GiniHealthSDKExample/GiniHealthSDKExample/AppCoordinator.swift
zladzeyka: I wanted to do it -> I'll comment out this line and leave a comment for it

HealthSDK/GiniHealthSDK/Sources/GiniHealthSDK/Core/PaymentReviewViewController.swift
zladzeyka: These lines will be called automatically when `selectedPaymentProvider` will be set                     `self?.selectedPaymentProvider = provider`

a-szotyori: Thanks, I understand the change now.

HealthSDK/GiniHealthSDKExample/GiniHealthSDKExample/AppCoordinator.swift
zladzeyka: I've decided to remove it 🙈

PR #54, opened 2022-03-14
Pia 2339 disabled state pay button

zladzeyka:
- Remove `tintColor` manipulation.
- Add a property for `payButtonDisabledTextColor `

### DIFF
--- a/HealthSDK/GiniHealthSDKExample/GiniHealthSDKExample/AppCoordinator.swift
+++ b/HealthSDK/GiniHealthSDKExample/GiniHealthSDKExample/AppCoordinator.swift
@@ -149,7 +149,6 @@ final class AppCoordinator: Coordinator {
         let regularFont = UIFont(name: "Avenir", size: 15) ?? UIFont.systemFont(ofSize: 15)
         configuration.customFont = GiniFont(regular: regularFont, bold: regularFont, light: regularFont, thin: regularFont)
         // Pay button configuration
-        configuration.payButtonTextColor = GiniColor(lightModeColor: .white, darkModeColor: .white)
         // Uncomment to test disabled state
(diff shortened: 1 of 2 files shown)


HealthSDK/GiniHealthSDK/Sources/GiniHealthSDK/Core/PaymentReviewViewController.swift
```diff
@@ -290,12 +290,14 @@ public final class PaymentReviewViewController: UIViewController, UIGestureRecog
...
         if let textHexColor = UIColor(hex: textColorString) {
-            payButton.tintColor = UIColor.from(giniColor: GiniColor(lightModeColor: textHexColor, darkModeColor: textHexColor))
+            payButton.textColor = UIColor.from(giniColor: GiniColor(lightModeColor: textHexColor, darkModeColor: textHexColor))
         }
     }
     fileprivate func configurePayButtonInitialState() {
         payButton.disabledBackgroundColor = UIColor.from(giniColor: giniHealthConfiguration.payButtonDisabledBackgroundColor)
         payButton.isEnabled = false
+        payButton.disabledTextColor = UIColor.from(giniColor: giniHealthConfiguration.payButtonDisabledTextColor)
+        payButton.textColor = UIColor.from(giniColor: giniHealthConfiguration.payButtonTextColor)
```
a-szotyori, line 300: Do we need `payButtonTextColor`? It will be overridden by the `PaymentProvider`.

a-szotyori (commented):
Looks good. I have only one question.

HealthSDK/GiniHealthSDK/Sources/GiniHealthSDK/Core/PaymentReviewViewController.swift
zladzeyka: Yes, the property `payButtonTextColor` will not be needed. Only if there's no payment provider as a default option.

PR #76, opened 2022-05-02
Pia 2485 pinning

zladzeyka:
In order to fix `#if can import` which is not working with SPM 🥲 and set `URLSessionDelegate` in `SessionManager`:
 - Add `URLSessionDelegate` class for passing into the lib with pinning initializer.
 - Add sessionDelegate in SessionManager initializer.
 - Add examples for `GiniBankAPILibraryPinning` and `GiniBankSDKPininng`.
PIA-2485


a-szotyori (changes requested):
Nice work! I left some minor comments.

a-szotyori (approved):

PR #81, opened 2022-05-05
Pia 2528 fix ra feedback sending

a-szotyori:
Introduces a breaking change by adding compound extractions to the `sendFeedback` method of the `GiniCaptureNetworkService`.

I had to also update our examples to fix feedback sending issues.


zladzeyka (approved):
🎉👏🏻👍🏻

PR #99, opened 2022-07-06
Pia 2730 disable return reasons

zladzeyka:
- Allows disabling return reasons for line items


BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/ReturnAssistant/DigitalInvoiceViewController.swift
```diff
@@ -380,10 +380,11 @@ extension DigitalInvoiceViewController: DigitalLineItemTableViewCellDelegate {
                 self.invoice?.lineItems[viewModel.index].selectedState = .deselected(reason: nil)
                 return
             }
-            
-            presentReturnReasonActionSheet(for: viewModel.index,
-                                           source: cell.modeSwitch,
-                                           with: returnReasons)
+            if returnAssistantConfiguration.enableReturnReasons {
```
a-szotyori, line 383: This should be moved into the `guard` at line 379. Otherwise the `selectedState` is not set to `.deselected(reason: nil)`.

a-szotyori (changes requested):
I found an itsy-bitsy 🐞.

a-szotyori (approved):

PR #107, opened 2022-07-14
Pia 2754-0-qantity

chriskryn:
- when the toggle will be deactivated the rest of the screen is turning grey
- fix unwanted behaviour when 0 quantity is pressed


zladzeyka (approved):
👍🏻👏🏻

PR #110, opened 2022-07-15
Pia 2787 limit quantity

chriskryn:
- limit quantity to 99999
- add margin between total price and label


zladzeyka (changes requested):
Looks good 👏🏻, 
only a 1 small renaming wish 😉

zladzeyka (approved):
👍🏻👏🏻🎉

PR #111, opened 2022-07-18
PIA-2781 pay button

chriskryn:
- Pay button is disabled when total amount is 0, irrelevant of how many items are selected.
- When button is disabled the label should be just “Pay”.
- the pay button is just 'Pay' when 0 items selected and there is a amount greater than 0


zladzeyka (approved):
🎉👏🏻😊Very nice!

PR #117, opened 2022-07-22
fix(GiniBankSDK):

zladzeyka:
Remove "Save functionality" when the back button is clicked.

PIA-2810

### DIFF
--- a/BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/ReturnAssistant/LineItemDetailsViewController.swift
+++ b/BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/ReturnAssistant/LineItemDetailsViewController.swift
@@ -75,12 +75,6 @@ class LineItemDetailsViewController: UIViewController {
         )
     }
     
-    override func viewWillDisappear(_ animated: Bool) {
-        super.viewWillDisappear(animated)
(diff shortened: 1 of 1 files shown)


BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/ReturnAssistant/LineItemDetailsViewController.swift
```diff
@@ -79,13 +79,6 @@ class LineItemDetailsViewController: UIViewController {
         super.viewWillDisappear(animated)
         guard let lineItem = lineItem, !lineItem.isUserInitiated else { return }
```
a-szotyori, line 80: This line has no use anymore. There method is empty after the guard so there is nothing to guard...

The whole viewWillDisappear method has no use anymore and can be deleted.

a-szotyori (changes requested):
You can delete more code. 😉

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/ReturnAssistant/LineItemDetailsViewController.swift
zladzeyka: Good catch 😉

a-szotyori (approved):

chriskryn (approved):
:+1:

PR #119, opened 2022-07-27
PIA-2257 xcframeworks script

chriskryn:
xcframeworks build automation script


.github/workflows/bank-sdk.build.xcframeworks.yml
```diff
@@ -0,0 +1,101 @@
+name: Build BankSDK XCFrameworks
+on:
+  push:
```
a-szotyori, line 3: We need also `workflow_call` as a trigger in `on` to make this callable from another workflow. https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow

a-szotyori (changes requested):
Wow, Awesome! There is only a small thing related to making the workflow callable from another workflow.

.github/workflows/bank-sdk.build.xcframeworks.yml
zladzeyka: like we have https://github.com/gini/gini-mobile-ios/blob/main/.github/workflows/bank-api-library.publish.docs.yml#L6

a-szotyori (approved):

zladzeyka (approved):
🎉👏🏻😊 Great job @chriskryn

PR #124, opened 2022-08-08
Pia 2865 fonts

zladzeyka:
GiniCaptureSDK, GiniBankSDK: 
- Adds default text style fonts,
- Allows customization of a specific text style.
- Add depreciation for old `customFont` property.

### DIFF
--- a/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/GiniConfiguration.swift
+++ b/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/GiniConfiguration.swift
@@ -83,10 +83,10 @@ import UIKit
         return CustomDocumentValidationResult.success()
     }
     
+    // MARK: - TODO DELETE
     /**
      Sets the font used in the GiniCapture library by default.
      */
-    @available(*, deprecated, message: "Please use `updateFont(font: UIFont, for textStyle: UIFont.TextStyle)` instead")
     @objc public lazy var customFont: GiniCaptureFont = GiniCaptureFont(regular: UIFont.systemFont(ofSize: 14,
                                                                                                  weight: .regular),
                                                                       bold: UIFont.systemFont(ofSize: 14,
--- a/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/GiniConfiguration.swift
+++ b/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/GiniConfiguration.swift
@@ -86,6 +86,7 @@ import UIKit
     /**
      Sets the font used in the GiniCapture library by default.
      */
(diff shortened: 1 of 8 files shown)


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/GiniConfiguration.swift
```diff
@@ -717,4 +718,31 @@ import UIKit
     @objc public var backToMenuButtonResource: PreferredButtonResource?
     @objc public var nextButtonResource: PreferredButtonResource?
     @objc public var cancelButtonResource: PreferredButtonResource?
+    
+    /**
+     Set dictionary of fonts for available text styles. Used internally.
+     */
+    public var textStyleFonts: [UIFont.TextStyle: UIFont] = [
```
a-szotyori, line 725: In Bank SDK's config this is private and here it is public. 🤔 I know that this makes it convenient to pass the font configuration from Bank SDK to Capture SDK, but it exposes internal concerns.

I would recommend to make this private so that the Capture SDK's public API only allows font configuration via `updateFont()`. Then in `GiniBankConfiguration` call `updateFont()` for each text style. It's much more verbose, but it keeps the public API clean.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/GiniConfiguration.swift
```diff
@@ -86,6 +86,7 @@ import UIKit
     /**
      Sets the font used in the GiniCapture library by default.
      */
+    @available(*, deprecated, message: "Please use `updateFont(font: UIFont, for textStyle: UIFont.TextStyle)` instead")
```
a-szotyori, line 89: What do you think about adding a todo here to remove this once it's not used anymore (after all screens were updated to the new UI)?

a-szotyori (commented):
Nice! 👍 There is one thing we need to discuss though. 🙂

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/GiniConfiguration.swift
zladzeyka: I was trying to do so but I was not able to iterate through UIFont.TextStyle 🤔. I'll check if I can do something else with it.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/GiniConfiguration.swift
zladzeyka: Ok, I'll add it.

a-szotyori (approved):

PR #135, opened 2022-08-25
Pia 2955 no result accessibility

chriskryn:


zladzeyka (commented):
Great work 👏🏻👍🏻🎉
Inspector didn't find any issues. The only thing I've noticed when I test `Voice over`: it reads only the 1st  visible cell for the table view, which means only one tip and no others, can we do something with it?

zladzeyka (approved):
We've tested 'voice-over' - it works as expected.

PR #143, opened 2022-09-07
Refactor ginicolor initializer

vizaknaidavid:
Refactored GiniColors initializer

Removed the long initilizer
Added util function to return UIColor for cleaner code
Remove UIColor extension which did the same as the new function, but less readable


zladzeyka (approved):
Looks 👍🏻👏🏻.

PR #150, opened 2022-09-12
Pia 3027 allow to add custom loading indicator with a custom animation

vizaknaidavid:
For testing it, you can uncomment the custom loading indicator configuration from the BankSDK example appcoordinator.

Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-09-12 at 15 00 06


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Analysis/AnalysisViewController.swift
```diff
@@ -244,6 +249,65 @@ import UIKit
...
+        addLoadingView(intoContainer: loadingIndicatorContainer)
+
+        if let loadingIndicator = giniConfiguration.analysisScreenLoadingIndicator {
+            addLoadingText(below: loadingIndicator.injectedView())
+            loadingIndicator.startAnimation()
+        } else {
+            addLoadingText(below: loadingIndicatorView)
+            loadingIndicatorView.startAnimating()
+        }
+    }
```
chriskryn, line 265: @vizaknaidavid @zladzeyka  Could you tell me how custom loading indicator should work? Should we allow displaying both at the same time ? Or one replaces another?

chriskryn (changes requested):
There are some linter issues in AnalysisViewController

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Analysis/AnalysisViewController.swift
zladzeyka: @chriskryn  there should be 1 loading indicator 😎: default or with a custom view.

zladzeyka: @vizaknaidavid there is no need to add a custom view for a loading indicator to a super view if giniConfiguration.analysisScreenLoadingIndicator is not set,

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Analysis/AnalysisViewController.swift
```diff
@@ -244,6 +249,65 @@ import UIKit
...
+        if let customLoadingIndicator = giniConfiguration.analysisScreenLoadingIndicator?.injectedView() {
+            loadingIndicator = customLoadingIndicator
+        } else {
+            loadingIndicator = loadingIndicatorView
+        }
+
+        if let container = container {
+            container.translatesAutoresizingMaskIntoConstraints = false
+            loadingIndicator.translatesAutoresizingMaskIntoConstraints = false
+            self.view.addSubview(container)
```
chriskryn, line 289: @vizaknaidavid  Could you remove the 'self' when it's not necessary?

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Analysis/AnalysisViewController.swift
chriskryn: @zladzeyka  Ok  :slightly_smiling_face: , I think it is used to center both indicator and text within the screen.

chriskryn: @vizaknaidavid @zladzeyka So do we need a text under custom indicator ? Then I would think that the custom indicator should be enough .

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Analysis/AnalysisScreenLoadingIndicatorAdapter.swift
```diff
@@ -0,0 +1,21 @@
...
+//  Created by David Vizaknai on 12.09.2022.
+//
+
+import Foundation
+/**
+*   Adapter for injecting a custom loading indicator for the analysis viewcontroller.
+*/
+public protocol AnalysisScreenLoadingIndicatorAdapter: InjectedViewAdapter {
+    /**
+     *  Called when the screen is loaded, can be also called from the AnalysisViewController's showAnimation() method.
```
a-szotyori, line 14: For me the intention of the documentation here is to inform clients about what should they do when implementing this method. The questions the documentation should answer is: "As a client what do I have to do in this method? When is this method called? Why do I have to implement it?"

We don't need to mention the `showAnimation()` method here. It is currently public, but we'll have to make it internal when we remove the Component API. Clients won't be able to interact directly with our screen view controllers anymore.

My change suggestion would be:
```suggestion
     *  Called when the screen is loaded. You should start the loading indicator animation in this method.
```

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Analysis/AnalysisScreenLoadingIndicatorAdapter.swift
```diff
@@ -0,0 +1,21 @@
...
+     */
+    func startAnimation()
+    /**
+     *  Can be called from the AnalysisViewController's hideAnimation() method.
```
a-szotyori, line 18: Similarly to the above comment my change proposal would be:
```suggestion
     *  Called when the screen has disappeared. You should stop the loading indicator animation in this method.
```

a-szotyori (changes requested):
Looks great. I only have comments regarding the documentation.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Analysis/AnalysisViewController.swift
vizaknaidavid: @zladzeyka I add a container that is centered in the screen in which I add the loading indicator and the text, so they are centered in the screen. If the custom indicator is not set, the conatiner will contain the default loading and the text.

vizaknaidavid: So can this thread be resolved without changes? @chriskryn  @zladzeyka  :D

zladzeyka: Yes, I think so ;)

chriskryn (approved):
Looks Good! :+1:

a-szotyori (approved):

PR #157, opened 2022-09-30
Pia 3163 update xcode

chriskryn:
change system type cell into custom cell in example apps. It seems it causes layout problems in executable version


a-szotyori (approved):

zladzeyka (approved):
👍🏻😊

PR #160, opened 2022-10-06
Pia 2949 bottom navigation bar no result screen

chriskryn:


zladzeyka (commented):
Looks good 😉👏🏻👍🏻
Default implementation works well. 
[Suggestion]@chriskryn could you please add some custom navigation adapter( maybe just reuse a default one with a little change)?
[Clarification] when I use `open with` with pdf from the Files folder `<` button closes the SDK, let's check with PO if it's the correct behavior.

https://user-images.githubusercontent.com/3911816/194593271-53051d83-eb0f-42e8-bf47-7c70c25a1d53.mp4

zladzeyka (approved):
Thank you 👏🏻👍🏻😊

PR #168, opened 2022-10-20
PIA-3213_qrcode_fix

zladzeyka:
- Fixed condition for showing popup for unsupported QR ( here IBAN was invalid)
 Use QR Codes for the test from [PIA-3213](https://ginis.atlassian.net/browse/PIA-3213)


a-szotyori (approved):
👏 Nice! Thank you for applying the suggestions!

chriskryn (approved):
Looks good!:+1:

PR #171, opened 2022-10-24
PIA-3218: fix(GiniCaptureSDK): fix back button on camera screen when bottom navigation bar enabled

chriskryn:
PIA-3218


zladzeyka (approved):
Great job!

PR #176, opened 2022-11-02
Pia 3209 camera setup issue

chriskryn:
- fix presence of activity indicator, completion handler executed after all session processing is finished (including qrcode session setup )


zladzeyka (approved):
Looks good! 😉

PR #186, opened 2022-11-10
Pia 2919 camera navigation

chriskryn:
- updated navigation and display of navigation buttons
- added  missing example implementation of bottom navigation bar


zladzeyka (approved):
😊👏🏻

PR #189, opened 2022-11-17
Pia 3307 Shorten button titles and group them in the middle

vizaknaidavid:
Simulator Screen Shot - iPhone 13 Pro - 2022-11-17 at 15 52 copy

Without and with bottom navigation


zladzeyka (approved):
Looks good!

PR #203, opened 2022-11-29
Pia 3333 error screen actions

chriskryn:


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/NoResult/NoResultScreenViewController.swift
```diff
@@ -93,7 +76,7 @@ final public class NoResultScreenViewController: UIViewController {
     private let tableRowHeight: CGFloat = 44
```
zladzeyka, line 76: [Suggestion] can we extract the magic numbers in the Constants?

zladzeyka (changes requested):
👏🏻😊👍🏻 Looks good! Thank you
I have 2 tiny suggestions.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Resources/de.lproj/Localizable.strings
```diff
@@ -130,6 +131,20 @@
 "ginicapture.noresult.title" = "TODO Fehler";
```
zladzeyka, line 131: [Suggestion] can we update `ginicapture.noresult.title` value as well?

zladzeyka (changes requested):
Good job😊 @chriskryn 👏🏻
I have only a few small suggestions 😇:
I've noticed that after we see the error dialog after we've captured 10 pages and back to the Review screen - the blue selection is gone.
For the bottom navigation in this branch, I've noticed that when we tap on `<`  we navigate to the Review screen.

zladzeyka (approved):
Looks 👍 👏🏻

zladzeyka (approved):

PR #208, opened 2022-12-02
Pia 3328 error screen topbar

chriskryn:
Top and bottom navigation bar


BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
```diff
@@ -280,22 +280,20 @@ extension GiniBankNetworkingScreenApiCoordinator {
...
-                
-                networkDelegate.displayError(withMessage: .localized(resource: AnalysisStrings.analysisErrorMessage),
-                                             andAction: {
-                                                 self.startAnalysisWithReturnAssistant(networkDelegate: networkDelegate)
-                                             })
-                 
+                DispatchQueue.main.async { [weak self] in
+                    guard error != .requestCancelled else { return }
+                    self?.displayError(errorType: ErrorType(error: error))
+
```
vizaknaidavid, line 286: [Suggestion] No need for empty line here

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Error/ErrorScreenViewController.swift
```diff
@@ -35,6 +30,17 @@ class ErrorScreenViewController: UIViewController {
             secondTitle: NSLocalizedStringPreferredFormat(
                 "ginicapture.error.backToCamera",
                 comment: "Back to camera"))
+        view.translatesAutoresizingMaskIntoConstraints = false
+        if viewModel.isEnterManuallyHidden() == false {
```
vizaknaidavid, line 34: `if viewModel.isEnterManuallyHidden() == false {
             view.enterButton.isHidden = false
         } else {
             view.enterButton.isHidden = true
         }`
         
         This whole block could be written like this:
         
         `view.enterButton.isHidden = viewModel.isEnterManuallyHidden()`
         
         Same for the one below

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Error/ErrorType.swift
```diff
@@ -0,0 +1,131 @@
...
+                }
+            } else {
+                self = .serverError
+            }
+        default:
+            self = .unexpected
+        }
+    }
+
+    func iconName() -> String {
```
vizaknaidavid, line 47: [Suggestion] Declaring these as variables not functions would reduce complexity

vizaknaidavid: `
var iconName: String {

        switch self {
        case .connection:
            return "errorCloud"
        case .request:
            return "alertTriangle"
        case .authentication:
            return "errorAuth"
        case .serverError:
            return "errorGlobe"
        case .unexpected:
            return "alertTriangle"
        case .uploadIssue:
            return "errorUpload"
        case .importError:
            return "alertTriangle"
        }
    }`

vizaknaidavid (changes requested):
Looks good, I just left some tiny suggestions

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Error/ErrorType.swift
chriskryn: hmm @vizaknaidavid  What kind of complexity ? I see that the difference would be just the way of invoking it without '()' parenthesis.

chriskryn:
Merged into PIA-3333 https://github.com/gini/gini-mobile-ios/pull/203

PR #213, opened 2022-12-08
Pia 2499 refactor how feedback is sent i os bank sdk capture sdk

vizaknaidavid:


zladzeyka (changes requested):
Thank you @vizaknaidavid! Look good. There are some things that I've noticed:

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
```diff
@@ -226,8 +226,6 @@ extension GiniBankNetworkingScreenApiCoordinator {
 
 
                 self.giniBankConfiguration.lineItems = result.lineItems
-
-                documentService.resetToInitialState()
```
zladzeyka: [Clarification] where is ` documentService.resetToInitialState()`called?

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Networking/GiniNetworkingScreenAPICoordinator.swift
```diff
@@ -125,11 +127,8 @@ import GiniBankAPILibrary
                 
                 
                 let result = AnalysisResult(extractions: extractions, lineItems: result.lineItems, images: images, document: document)
-                
-                let documentService = self.documentService
-                
+                                
                 self.resultsDelegate?.giniCaptureAnalysisDidFinishWith(result: result)
-                documentService.resetToInitialState()
```
zladzeyka: [Clarification] ` documentService.resetToInitialState()` needs to be called after the feedback was sent. Where do we call it?

zladzeyka (changes requested):
Great progress 🎉👏🏻

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
vizaknaidavid: As you suggested:
`documentService.resetToInitialState() needs to be moved at the end of your cleanup method.`

BankSDK/GiniBankSDKExample/Tests/ExtractionFeedbackIntegrationTest.swift
```diff
@@ -49,7 +46,7 @@ class ExtractionFeedbackIntegrationTest: XCTestCase {
         // (subsequent steps are in CaptureResultsDelegateForExtractionsFeedbackTest)
         self.getExtractionsFromGiniBankSDK(delegate: CaptureResultsDelegateForExtractionsFeedbackTest(integrationTest: self, expect: expect))
 
-        wait(for: [expect], timeout: 30)
+        wait(for: [expect], timeout: 60)
```
zladzeyka, line 49: Let's keep 30 sec - it should be enough or maybe we have a special reason for 60 sec?

zladzeyka (approved):
Looks good!😊 Great job!
I only notice that timeout increased up to 60 sec - we can keep 30, it should be enough.

BankSDK/GiniBankSDKExample/Tests/ExtractionFeedbackIntegrationTest.swift
vizaknaidavid: With 30 seconds, locally the success rate was not 100%, with this increase it succeeded all the time. But we still have the problem with the CI, so maybe this is not a valid fix and I have to check the problem elsewhere

PR #216, opened 2022-12-14
Pia 3347 increase cropped area by 15 on each side

vizaknaidavid:
Crop area should be bigger than the white rectangle with 15%, but only if the image size allows it.


zladzeyka (approved):
Great 😊👏🏻
I have only 1 small clarification question 😉.

PR #217, opened 2022-12-15
fix(GiniCaptureSDK): Remove public access modifier

vizaknaidavid:
PIA-2954


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/NoResult/BottomButtonsViewModel.swift
```diff
@@ -8,12 +8,12 @@
 
 import Foundation
 
-public final class BottomButtonsViewModel {
```
zladzeyka, line 11: [Sugesstion] I am afraid that we would need to keep it public. This class for the model will be reused in error screen as well 🧐

zladzeyka (approved):
Looks good to me. I would suggest merging it after the error screen and removing component API PRs 😊

PR #218, opened 2022-12-16
Fix QR code scanning only if flag of QR reading is disabled

vizaknaidavid:
If you enable QRCode Scan ONLY but disable QR code scanning, nothing should happen.


zladzeyka (approved):
Looks good 👍🏻

PR #225, opened 2023-01-04
fix(GiniCaptureSDK): Fix button allignment on review screen

vizaknaidavid:
Navigation is already fixed, in this PR I fixed the button layout if multipage is disabled


zladzeyka (approved):
Looks good 🎉😊

PR #228, opened 2023-01-05
Documentation component api

vizaknaidavid:


BankSDK/GiniBankSDK/Documentation/source/QR Code scanning guide.md
```diff
@@ -9,13 +9,23 @@ Enable QR code scanning
...
 ```swift
-let giniBankConfiguration = GiniBankConfiguration()
+let giniBankConfiguration = GiniBankConfiguration.shared
 ...
 ...
 ...		
 giniBankConfiguration.qrCodeScanningEnabled = true
 ```
 
+For allowing to scan only QR code without the ability to take images, you need to enable the flag for this in the `GiniBankConfiguration`, but also the QR scanning ability, like so:
```
zladzeyka, line 19:
```suggestion
For activating QR code-only mode without the ability to take images, you need to enable the flag in the `GiniBankConfiguration`, but also the QR scanning ability, like so: 
```

zladzeyka (commented):
Look great! I have only a small suggestion 😇

PR #236, opened 2023-01-12
Pia 3478 cleanup gini configuration and gini bank configuration options

vizaknaidavid:


BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankConfiguration.swift
```diff
@@ -248,19 +248,6 @@ public final class GiniBankConfiguration: NSObject {
      Sets the continue button text in the navigation bar on the review screen.
      */
     @objc public var navigationBarReviewTitleContinueButton = ""
```
zladzeyka, line 250: I think navigationBarReviewTitleContinueButton can be removed as well 😉

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Resources/en.lproj/Localizable.strings
```diff
@@ -81,11 +81,7 @@
 "ginicapture.onboarding.qrCode.title"  = "QR codes supported";
 "ginicapture.onboarding.qrCode.description"  = "Besides invoices, QR codes
 are also supported";
-"ginicapture.review.top"                = "Is the document complete, in focus and in the correct orientation?";
-"ginicapture.review.bottom"             = "Double-tap to check the focus";
-"ginicapture.review.rotateButton"       = "Rotate document";
 "ginicapture.review.documentImageTitle" = "Document";
-"ginicapture.review.unknownError"       = "An unknown error has occurred on the review screen";
```
zladzeyka, line 88: [Suggestion] You need to return back
` "ginicapture.review.unknownError" = "An unknown error has occurred on the review screen";
` 
It stays in German translation and is used above.

zladzeyka (changes requested):
Great 👍 😊

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankConfiguration.swift
vizaknaidavid: I will check the button resources that we have on ScreenAPICoordinator, and if they are not used, I will remove all of them

vizaknaidavid: It seems that they are used in Xamarin code, so it should stay :D

zladzeyka (approved):
👏🏻

PR #239, opened 2023-01-13
Pia 3490 i os bottom nav missing on image picker

vizaknaidavid:
Screenshots from iPhone 13 pro, iOS 15 and iPhone SE iOS 12.4

pr


zladzeyka (commented):
Great job!
We need to add the example implementation as well 😉

zladzeyka (approved):
Looks good 😉

PR #240, opened 2023-01-17
fix(GiniCaptureSDK): Fix flash toggle apperance

vizaknaidavid:
Fix flash toggle hiding when flag is set from config. 
Please test it on a device with no flash. I don't have any right now.


zladzeyka (approved):
Looks good to me 👍🏻

PR #267, opened 2023-02-09
Pia 3716 camera error dialog customization options are missing

vizaknaidavid:
Added customization options and removed duplicate code


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Extensions/UIViewController.swift
```diff
@@ -78,7 +85,18 @@ extension UIViewController {
                                  confirmActionTitle: String? = nil,
                                  confirmAction: (() -> Void)? = nil) -> UIAlertController {
         
-        let alertViewController = UIAlertController(title: nil, message: message, preferredStyle: .alert)
+        let alertViewController = UIAlertController(title: nil, message: nil, preferredStyle: .alert)
+
+
+        let messageAttributes = [NSAttributedString.Key.font : UIFont.systemFont(ofSize: 17,
```
a-szotyori, line 91: [Suggestion] I think we shouldn't customise the font and text color. The background color is set by the system and it might be that the dark1 or light1 colors might not be properly visible. Here as in the file import dialog we should stick to the system defaults.

I would only keep the tintColor customisation.

a-szotyori (approved):

PR #277, opened 2023-02-21
Pia 3613 allow injecting a bottom navigation bar help

vizaknaidavid:
Simulator Screen Shot - iPhone 13 Pro - 2023-02-21 at 17 35 25
Simulator Screen Shot - iPhone 13 Pro - 2023-02-21 at 17 35 27

### DIFF
--- a/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Extensions/UIView.swift
+++ b/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Extensions/UIView.swift
@@ -6,7 +6,7 @@
 //
 
 import UIKit
-public extension UIView {
+extension UIView {
     /** Loads instance from nib with the same name. */
     func loadNib() -> UIView {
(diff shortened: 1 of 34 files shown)


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Extensions/UIView.swift
```diff
@@ -6,7 +6,7 @@
 //
 
 import UIKit
-extension UIView {
+public extension UIView {
```
a-szotyori, line 9: [Request] I am worried making extensions to UIView public. It might clash with client app's UIView extensions. Also the method `loadNib()` uses `giniCaptureBundle()`. 

As these are helper convenience methods I would rather we duplicate the methods that are needed in Bank SDK.

What do you think? Is there a better option?

a-szotyori (changes requested):
Just minor things.

a-szotyori (approved):
Thank you!

PR #280, opened 2023-02-23
Pia 3162 expose extraction candidates

a-szotyori:
I exposed a field through `AnalysisResult` to allow clients to get the extraction candidates.

Extraction candidates contain all identified extractions, for example if the invoice has three IBANs all of them will be added to the extraction candidates. One of them is picked as the main extraction, but clients may want to use one of the candidates instead.


vizaknaidavid (approved):
Looks good!

PR #284, opened 2023-02-28
Pia 3649 switch to ad hoc distribution

a-szotyori:


vizaknaidavid (approved):
Looks good, works ok!

PR #290, opened 2023-03-02
feature(GiniBankSDK): Add bottom navigation for return assistant overview

vizaknaidavid:
bottomnav

### DIFF
--- a/BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/ReturnAssistant/DigitalInvoice/DigitalInvoiceOverviewNavigationBarBottomAdapter.swift
+++ b/BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/ReturnAssistant/DigitalInvoice/DigitalInvoiceOverviewNavigationBarBottomAdapter.swift
@@ -43,7 +43,7 @@ public protocol DigitalInvoiceOverviewNavigationBarBottomAdapter: InjectedViewAd
     func updateButtonState(enalbed: Bool)
 
     /**
-     *  Lays out the views based on the table view's position in the return assistant overview screen
+     *  Lays out the views based on the table view's position in the return assistant overview screen. Use this to align views with the table view
      *
      * - Parameter view: The tableview on the return assistant overview screen
      */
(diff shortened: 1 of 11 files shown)


BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/ReturnAssistant/DigitalInvoice/DigitalInvoiceOverviewNavigationBarBottomAdapter.swift
```diff
@@ -0,0 +1,51 @@
...
+     * - Parameter callback: An  action callback, which should be retained and called in help button action method
+     */
+    func setHelpButtonClickedActionCallback(_ callback: @escaping  () -> Void)
+
+    /**
+     *  Set the total price on the bottom navigation bar. Called when the total price changes
+     *
+     * - Parameter price: A string which contains the currency and the price
+     */
+    func updateTotalPrice(with price: String?)
```
a-szotyori, line 36: [Request] `updateTotalPrice(priceWithCurrencySymbol price:)` to make it obvious that the price includes the currency symbol.

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/ReturnAssistant/DigitalInvoice/DigitalInvoiceOverviewNavigationBarBottomAdapter.swift
```diff
@@ -0,0 +1,51 @@
...
+     * - Parameter enabled: A bool value to reflect the state of the button
+     */
+    func updateButtonState(enalbed: Bool)
+
+    /**
+     *  Lays out the views based on the table view's position in the return assistant overview screen. Use this to align views with the table view
+     *
+     * - Parameter view: The tableview on the return assistant overview screen
+     */
+    func setupViewsRelated(to view: UIView)
```
a-szotyori, line 50: [Suggestion] I would prefer keeping the adapter independent from the our requirements of the screen's layout. The need to align the width with the table view is our requirement, but clients might want to layout their bottom navbar differently. By adding `setupViewsRelated(to:)` to the protocol we make it a requirement for all clients.

Also on Android we don't have this method and would be great if we could align the adapters as much as possible.

Is it possible to add the same margin constraints that the table view has to our default implementation so that we don't need `setupViewsRelated(to:)`? Alternatively is it possible to have `setupViewsRelated(to:)` only in our default implementation and then invoke it only after checking that it's available (with somekind of "responds to" check)?

a-szotyori (changes requested):
Some minor stuff.

PR #304, opened 2023-03-20
Pia 3600 update reference documentation

vizaknaidavid:
Removed public modifiers where there were no need for them


zladzeyka (approved):
👍

PR #310, opened 2023-03-22
Pia 3889 update the bottom navigation bar buttons

vizaknaidavid:
Navigation buttons on the bottom navigation bar should not scale.


zladzeyka (approved):
👍

PR #332, opened 2023-04-07
Hide keyboard when tapping outside of edit container - QA

vizaknaidavid:
Fix animation when keyboard is presented.


zladzeyka (approved):
Looks good ;)

PR #339, opened 2023-04-11
chore(GiniCaptureSDK): Centre the header for the no results and error screen

zladzeyka:
Centre the header for the no results and error screen

PIA-3945


vizaknaidavid:
image

On tablet it looks okay, but on phone the icon is no longer aligned with the edge of the tableview and if the text is bigger, it will grow outside of the header and will not be visible.

vizaknaidavid (approved):
Looks good now 🥳

PR #347, opened 2023-04-17
Pia 3986 allow path in custom base url

a-szotyori:
Changes for version 1.x.

Example of how to set the path when using the Bank SDK:
```swift
    let viewController = GiniBank.viewController(withClient: client,
                                                 configuration: configuration,
                                                 resultsDelegate: self,
                                                 api: .custom(domain: "scan2bank-service-v1-scan2bank-itu.apps.caas-ekad-test01.rz.bankenit.de",
                                                              path: "/rest/de.fiduciagad.scan2bank.api.v1.Scan2BankApi"), 
                                                              tokenSource: nil)
```

### DIFF
--- a/.github/workflows/bank-api-library.build.docs.yml
+++ b/.github/workflows/bank-api-library.build.docs.yml
@@ -12,11 +12,11 @@ on:
   
 jobs:
   documentation:
-    runs-on: macOS-11    
+    runs-on: macOS-12
     steps:
     - uses: maxim-lobanov/setup-xcode@v1
       with:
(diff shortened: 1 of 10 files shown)


zladzeyka (commented):
Looks good 🥳,
Only one small question do we need some changes for Pinning as well?

zladzeyka (approved):

a-szotyori:
Pinning doesn't need changes, because it also uses the `APIDomain` enum and `Resource` protocol from `GiniBankAPILibrary`.

PR #355, opened 2023-04-21
Pia 3914 release candidate ios window about import is in german language but locale of device is english

vizaknaidavid:


zladzeyka (approved):
Great job  😇

PR #379, opened 2023-06-14
fix(GiniCaptureSDK): Fix the display of the error screen on the main thread after the server returned the error

zladzeyka:
Fix the display of the error screen on the main thread after the server returned the error

PIA-4271

### DIFF
--- a/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Networking/GiniNetworkingScreenAPICoordinator.swift
+++ b/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Networking/GiniNetworkingScreenAPICoordinator.swift
@@ -139,7 +139,9 @@ extension GiniNetworkingScreenAPICoordinator {
                 self.deliver(result: extractions, and: self.documentService.document, to: networkDelegate)
             case .failure(let error):
                 guard error != .requestCancelled else { return }
-                networkDelegate.displayError(errorType: ErrorType(error: error), animated: true)
+                DispatchQueue.main.async {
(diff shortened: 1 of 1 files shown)


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Networking/GiniNetworkingScreenAPICoordinator.swift
```diff
@@ -134,12 +134,14 @@ import GiniBankAPILibrary
 extension GiniNetworkingScreenAPICoordinator {
     fileprivate func startAnalysis(networkDelegate: GiniCaptureNetworkDelegate) {
         self.documentService.startAnalysis { result in
-            switch result {
-            case .success(let extractions):
-                self.deliver(result: extractions, and: self.documentService.document, to: networkDelegate)
-            case .failure(let error):
-                guard error != .requestCancelled else { return }
-                networkDelegate.displayError(errorType: ErrorType(error: error), animated: true)
+            DispatchQueue.main.async {
```
ValentinaIancu-Gini, line 137: It seems that also `self.deliver(result: extractions, and: self.documentService.document, to: networkDelegate)` is switching to the main thread and this can potentially lead to recursive or nested asynchronous calls. This can cause unintended consequences, such as excessive resource consumption or unexpected order of execution. 
Having two `DispatchQueue.main.async` calls does not inherently cause a problem. However, it's important to consider the order and timing of the code execution to avoid unexpected behavior.

ValentinaIancu-Gini (approved):

PR #388, opened 2023-07-12
PIA- 4345 add entry point for jpeg images

zladzeyka:


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/GiniConfiguration.swift
```diff
@@ -558,4 +558,17 @@ import GiniBankAPILibrary
...
+    /**
+     Enum that represents entry points for the Gini Capture SDK
+    */
+    public enum GiniEntryPoint: Int {
+        case button
+        case field
+    }
+
+    /**
+     Set the entry point for the Gini Capture SDK with a default value `GiniEntryPoint.button`
```
a-szotyori, line 571: [Suggestion]
```suggestion
     Set the entry point used for launching the Gini Capture SDK. 
     Default value is `GiniEntryPoint.button`.
```

CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/ImageMetaInformationManagerTests.swift
```diff
@@ -62,5 +62,42 @@ final class ImageMetaInformationManagerTests: XCTestCase {
...
+        let existingUserComment = metaInformation?.getMetaInformation(forKey: kCGImagePropertyExifUserComment as String)
+        let components = existingUserComment?.components(separatedBy: ",")
+        let userCommentComponent = components?.filter({ (component) -> Bool in
+            return component.contains("EntryPoint")
+        })
+        let equasionComponents = userCommentComponent?.last?.components(separatedBy: "=")
+        let entryPoint = equasionComponents?.last
+        XCTAssert(entryPoint == value,
+                  "failed to set value for entry point correctly on meta information")
+    }
```
a-szotyori, line 101: [Request] The first two tests test that you can set and read the "make" exif property (`kCGImagePropertyTIFFMake`) with the values "field" and "button". That's not in line with the test method names and also doesn't exercise the code lines you added. So the first two tests you can delete.

The third test does use the code lines you added and I would like to ask you duplicate it for testing the button entry point, too. You could call it `testButtonEntryPointToMetaInformation`.

CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/ImageMetaInformationManagerTests.swift
```diff
@@ -62,5 +62,42 @@ final class ImageMetaInformationManagerTests: XCTestCase {
...
```
a-szotyori, line 102: [Request] Would be great if you could add two tests (one for button and one for field entry point) with which you check that the "EntryPoint" is added to jpegs when using `imageByAddingMetadata()`. This is important I think because this is how we add "EntryPoint" to jpegs from the camera or using upload.

a-szotyori (changes requested):
Looks good overall! Just minor stuff plus test improvement ideas.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/ImageMetaInformationManager.swift
```diff
@@ -231,16 +232,29 @@ final class ImageMetaInformationManager {
         return nil
     }
 
+    fileprivate func entryFieldString(_ entryPoint: GiniConfiguration.GiniEntryPoint) -> String {
+            var stringValue: String
```
ValentinaIancu-Gini, line 236: [suggestion]: There is any reason why we can not do it directly in GiniConfiguration something like this:
```
public enum GiniEntryPoint: String {
        case button
        case field
    }
```
instead of declaring the enum of type Int and having this extra method for getting the string value of each case.

a-szotyori: I was also thinking about this for Android. I decided to do it like Nadya and separated enum values from their exif string encoding. If we use the enum value names directly in the exif, then if we change the enum values we also directly affect the exif and this would also not be obvious in the code.

If we keep enum values and their exif string encoding separate we avoid accidentally changing what we write into the exif. This is especially important because the analytics requies specific values in the exif.

ValentinaIancu-Gini: But on iOS we can do like this: 
```
public enum GiniEntryPoint: String {
        case button = "button"
        case field = "field" 
    }
```
or we can have a computed property for the `stringValue` like this:
```
public enum GiniEntryPoint {
        case button
        case field

var stringExifValue: String {
      switch self {
            case .button:
                return "button"
            case .field:
                return "field"
            }
      } 
    }
```
Any thoughts?

zladzeyka: @a-szotyori thanks for the comment :) you were very fast.

a-szotyori: @ValentinaIancu-Gini If the string value is important for exif, then I think it should stay in the class dealing with exif. In other words the string value of the enum is a concern for exif not for the enum therefore it should be closer to exif than to the enum.

ValentinaIancu-Gini: Ok, got it.
[suggestion]: we can rewrite the method to be more clear, efficient and short :
```
	fileprivate func entryFieldString(_ entryPoint: GiniConfiguration.GiniEntryPoint) -> String {
		switch entryPoint {
		case .button:
			return "button"
		case .field:
			return "field"
		}
	}
```

a-szotyori (approved):

PR #394, opened 2023-07-28
PIA-4391_fix_delete_icon_interaction

zladzeyka:
- Fix the `delete button`  interaction by adding all cell subviews to its `contentView` on the `Review screen`
 - Fix the subviews adding to its `contentView`  for cells on the `Help menu screen`, `Gallery screen`, inside capture suggestion cell - used in `No results` for image analysis, `Help Tips screen`.


ValentinaIancu-Gini:
@zladzeyka I think we also need to add the custom views inside the contentView in `CaptureSuggestionsCollectionHeader`

ValentinaIancu-Gini:
Also, all the constraints should be added between custom views and the content view not self. We need to change this in all the cells where we modified where we are adding the custom views.

zladzeyka:
> @zladzeyka I think we also need to add the custom views inside the contentView in `CaptureSuggestionsCollectionHeader`

CaptureSuggestionsCollectionHeader : UICollectionReusableView : UIView -> It doesn't have a contentView

ValentinaIancu-Gini:
> > @zladzeyka I think we also need to add the custom views inside the contentView in `CaptureSuggestionsCollectionHeader`
> 
> CaptureSuggestionsCollectionHeader : UICollectionReusableView : UIView -> It doesn't have a contentView

Pff I was checking apple documentation for UICollectionReusableView and I've scrolled down and saw a contentview but not I saw that is from UICollectionCell and not from UICollectionReusableView...sorry for the confusion.

zladzeyka:
> Also, all the constraints should be added between custom views and the content view not self. We need to change this in all the cells where we modified how we are adding the custom views.

Good point, I've updated it.

ValentinaIancu-Gini (approved):

ValentinaIancu-Gini:
@zladzeyka I fixed all the suggestions and comments :D.

a-szotyori (approved):
Looks good to me!

PR #415, opened 2023-09-08
fix(GiniCaptureSDK): Fix retry event tracking on the Error screen

zladzeyka:
PIA-4460


a-szotyori (approved):
I also tested it and I get the retry event now from the error screen.

PR #417, opened 2023-09-15
fix(GiniCaptureSDK): Fix color customisation to be applied for navigation left and right buttons

ValentinaIancu-Gini:
- used `tintedImageWithColor` method for the `back` button item

PIA-4413


zladzeyka (approved):

zladzeyka (approved):
Looks good!

PR #422, opened 2023-09-27
Pia 4356 on device iban detection

zladzeyka:


zladzeyka:
Changes implemented in https://github.com/gini/gini-mobile-ios/pull/427

PR #429, opened 2023-10-30
feat(GiniCaptureSDK): Limit IBAN detection to the camera frame

ValentinaIancu-Gini:
- display detected IBANs that are fully inside the frame

PIA-4618

### DIFF
--- a/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/CameraPreviewViewController.swift
+++ b/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/CameraPreviewViewController.swift
@@ -65,13 +65,6 @@ final class CameraPreviewViewController: UIViewController {
         return giniConfiguration.qrCodeScanningEnabled && giniConfiguration.onlyQRCodeScanningEnabled
     }()
 
-    // Transform from UI orientation to buffer orientation.
-    private var uiRotationTransform = CGAffineTransform.identity
-    // Transform bottom-left coordinates to top-left.
-    private var bottomToTopTransform = CGAffineTransform(scaleX: 1, y: -1).translatedBy(x: 0, y: -1)
-    // Transform coordinates in ROI to global coordinates (still normalized).
-    private var roiToGlobalTransform = CGAffineTransform.identity
-
     private var notAuthorizedView: UIView?
     private let giniConfiguration: GiniConfiguration
     private typealias FocusIndicator = UIImageView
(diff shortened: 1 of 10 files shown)


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/CameraPreviewViewController.swift
```diff
@@ -157,6 +149,19 @@ final class CameraPreviewViewController: UIViewController {
         setupConstraints()
     }
 
+    override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+
+        if isViewLoaded {
```
zladzeyka, line 155: [Suggestion] Let's extract the frame magic into a separate method and leave a comment there.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/CameraPreviewViewController.swift
```diff
@@ -366,6 +370,48 @@ final class CameraPreviewViewController: UIViewController {
...
+        var roiToGlobalTransform = CGAffineTransform.identity
+        // Region of video data output buffer that recognition should be run on.
+        // Gets recalculated once the bounds of the preview layer are known.
+        let roi = CGRect(x: 0, y: 0, width: 1, height: 1)
+        // Recalculate the affine transform between Vision coordinates and AVF coordinates.
+        roiToGlobalTransform = CGAffineTransform(translationX: roi.origin.x, y: roi.origin.y)
+            .scaledBy(x: roi.width, y: roi.height)
+
+        // Compensate for orientation (buffers always come in the same orientation).
+        switch currentOrientation {
```
zladzeyka, line 396: [Suggestion]: Could you please extract switch case into a separate method?

zladzeyka (commented):
Looks great @ValentinaIancu-Gini🥳
I have only 2 small suggestion about extracting code into separate methods.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/CameraPreviewViewController.swift
ValentinaIancu-Gini: I don't see the purpose in extracting it outside of this method since it sets some properties that are important just in this method and not in any other part. 
Also, that is the purpose of this method to set the correct text orientation and do the transformation.

Second thing is that I will prefer to keep this code in one method since for me is not very clear what are all the transformations done here and not spread it everywhere.

Also, I'm planning to check all the orientation properties that we set in this class when I will work to fix the bug for iPad(https://ginis.atlassian.net/browse/PIA-4627).

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/CameraPreviewViewController.swift
ValentinaIancu-Gini: Done

zladzeyka (approved):
😇

zladzeyka (approved):

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/CameraPreviewViewController.swift
zladzeyka: Agree!

PR #434, opened 2023-11-15
Bsdk 94 make flash off by default

ValentinaIancu-Gini:
- make `flashOnByDefault` disabled by default when the SDK starts
- adjust the name of the two settings options in the Settings screen related to flash

BSDK-94


zladzeyka (approved):
Thank you @ValentinaIancu-Gini 👏🏻. Looks good!

PR #440, opened 2023-12-11
fix(GiniBankSDK): Fix the displaying of the digital invoice onboarding screen

ValentinaIancu-Gini:
- RA flow -> after coming back from the help screen, the digital invoice onboarding screen should not be displayed

BSDK-222


zladzeyka (approved):
👏🏻 Great job

PR #441, opened 2023-12-13
fix(GiniBankSDK): Fix UI issues in Digital Invoice screen after disabling one item

ValentinaIancu-Gini:
- rounded corners and the separator visibility need to be reset to their default values when reusing cells after the call of `tableView.reloadData()`

BSDK-163


a-szotyori (approved):
Nice!

PR #446, opened 2024-01-19
Bsdk 27 anonymous user not logged in again after each ios gini bank api library launch

ValentinaIancu-Gini:
- Fixes anonymous user to log in again after each GiniBankApiLibrary launch
- cleanup code


ValentinaIancu-Gini:
@a-szotyori @zladzeyka I wasn't sure if I needed to add 
` try keychainStore.remove(service: .auth, key: .clientAccessToken)` in `removeStoredCredentials` method from `GiniBankAPI` class. I thought that was not necessary since `clientAccessToken` is the first token that we need to create a new user.
What do you think?

zladzeyka (approved):
😊 Great success!

PR #447, opened 2024-01-22
Bsdk 26 anonymous user not logged in again after each ios gini health api library launch

ValentinaIancu-Gini:
- Fixes anonymous user to log in again after each GiniHealthApiLibrary launch
- cleanup code

BSDK-26


zladzeyka (approved):
Great! Thank you 😇

PR #452, opened 2024-02-06
Publish health example

zladzeyka:
Added workflow for publishing Gini Health SDK Example

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -343,8 +343,13 @@ platform :ios do
     Setup Manual Signing for project at path
   DOC
   lane :setup_manual_signing do |options|
-    (project_path, team_id, target, target_extension, bundle_identifier, bundle_identifier_extension, profile_name, profile_name_extension, entitlements_file_path, entitlements_file_path_extension, ci) = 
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -356,6 +361,7 @@ platform :ios do
       entitlements_file_path: entitlements_file_path
     )
 
+    if (target_extension || bundle_identifier_extension || profile_name_extension || entitlements_file_path_extension)
```
a-szotyori, line 364: [Suggestion] I think by using OR (`||`), the statement in the `if` will be true even if some of the options are missing. Have you tried for example to pass to the lane only `target_extension` to see whether the `if` body is executed?

I am pretty sure that AND (`&&`) should be used, but Ruby is special... 🤔

a-szotyori (commented):
Nice! I left a suggestion.

fastlane/Fastfile
zladzeyka: and I was thinking it will fail if only `target_extension` is passed and you need to fix it by passing the other parameters? 🤔

a-szotyori (approved):

ValentinaIancu-Gini (approved):
Nice @zladzeyka !

PR #454, opened 2024-02-13
fix(GiniHealthAPILibraryPinning): Fixes error type with a wrong pinning certificate

zladzeyka:
fix(GiniHealthAPILibraryPinning): Fixes error type with a wrong pinning certificate


ValentinaIancu-Gini (approved):
Great job!!!! so happy that is fixed now!!!

PR #457, opened 2024-02-20
PP 67 Mapping of error messages with api defined status codes

ValentinaIancu-Gini:
Fix the error type when fetching the user access token so the error is not `unauthorized` one and tests


zladzeyka (approved):
👍🏻

PR #493, opened 2024-04-30
Pm 81 fix for custom api domain

a-szotyori:
While testing the Health API on stage I discovered that some requests were always using the default api domain instead of the custom one and so I had to apply these small fixes to be able to test on stage.


ValentinaIancu-Gini (approved):
Thank you @a-szotyori !

PR #494, opened 2024-05-03
PP 422 tapping the help button multiple times before the camera is active help screen is pushed multiple times

ValentinaIancu-Gini:


zladzeyka (approved):
🥳

PR #505, opened 2024-05-17
PP 390 Add missing event property to `screen_shown` event in analysis screen

ValentinaIancu-Gini:
- add `documentType` event property PP-390


mrkulik:
@ValentinaIancu-Gini  great catch of missing params🙌

mrkulik (approved):

PR #508, opened 2024-05-28
Pp 386 add user properties

mrkulik:
1. Mixpanel user properties added:
```
- entry_point
- gini_client_id
- is_voice_over_running
- is_guided_access_enabled
- is_bold_text_enabled
- is_grayscale_enabled
- is_speak_selection_enabled
- is_speak_screen_enabled
- is_assistive_touch_running
- return_reasons_enabled
- return_assistant_enabled
```
2. Change AnalyticsManager init place
3. Add EU server for MixPanel


BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
```diff
@@ -151,7 +154,9 @@ open class GiniBankNetworkingScreenApiCoordinator: GiniScreenAPICoordinator, Gin
         giniBankConfiguration = configuration
         giniBankConfiguration.documentService = documentService
         GiniBank.setConfiguration(configuration)
-
+        // TODO: No clientID user property for custom networking init
+        AnalyticsManager.userProperties = [.returnAssistantEnabled: configuration.returnAssistantEnabled,
```
ValentinaIancu-Gini, line 158: [suggestion]: I suggest having a separate method to set ` [.returnAssistantEnabled: configuration.returnAssistantEnabled,
                                           .returnReasonsEnabled: configuration.enableReturnReasons]` and reuse it.
 Maybe we would need to do same thing for client_id or not

ValentinaIancu-Gini (commented):
@mrkulik  Thank you for your good work. 

I've left some suggestions.

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
ValentinaIancu-Gini: No. I was thinking to have a method here to set those properties and use it in both place: line 135 and here line 158.

ValentinaIancu-Gini (approved):
@mrkulik great job!
I think the `entry_point' event or property should go in a separate pr since it might be a separate event.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Tracking/Models/EntryPointAnalytics.swift
```diff
@@ -0,0 +1,22 @@
+//
+//  AnalyticsEntryPoint.swift
```
ValentinaIancu-Gini, line 2: [suggestion] @mrkulik please also rename `AnalyticsEntryPoint` into `EntryPointAnalytics`

ValentinaIancu-Gini (approved):
Eveything looks good. Great job!

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Tracking/AnalyticsManager.swift
```diff
@@ -54,31 +61,61 @@ public class AnalyticsManager {
...
 
-            if let propertyValue = propertyValue as? [String] {
-                eventProperties[property.key.rawValue] = arrayToString(from: propertyValue)
+            mixpanelInstance?.registerSuperPropertiesOnce(superProperties)
+        } else {
+            for (property, value) in properties {
+                superProperties[property] = value
             }
         }
+    }
```
ValentinaIancu-Gini, line 100: [suggestion]: I suggest extracting common code into a method like this `mapPropertiesToDictionary`
```
private static func mapPropertiesToDictionary(_ properties: [AnalyticsPropertyType: AnalyticsPropertyValue]) -> [String: String] {
    var dictionary: [String: String] = [:]

    for (property, value) in properties {
        dictionary[property.rawValue] = convertPropertyValueToString(value)
    }

    return dictionary
}
```
and reuse it in both `trackUserProperties` and `trackUserProperties`.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Tracking/AnalyticsManager.swift
```diff
@@ -54,31 +61,61 @@ public class AnalyticsManager {
...
             }
 
-            if let propertyValue = propertyValue as? String {
-                eventProperties[property.key.rawValue] = propertyValue
+            mixpanelInstance?.people.set(properties: propertiesToTrack)
+        } else {
+            for (property, value) in properties {
+                userProperties[property] = value
             }
+        }
```
ValentinaIancu-Gini, line 83: [clarification]: I don't understand if `trackUserProperties` and `registerSuperPropertiesOnce` are called only once when the SDK is initialized, does it still make sense to have the else statement here? Also this methods are called right after we create a `mixpanelInstance`.
Also, we don't do anything when we track a normal event: `mixpanelInstance?.track(event: event.rawValue, properties: eventProperties)`

ValentinaIancu-Gini (commented):
@mrkulik Thank you updating this pr.
I still left some a suggestion and a clarification question.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Tracking/AnalyticsManager.swift
mrkulik: `trackUserProperties` and `registerSuperPropertiesOnce` can be called before Analytics SDK inits, because some of user properties (returnAssistantEnabled & returnReasonsEnabled) and super properties (client id) are tracked from BankSDK. I made arrays private and all tracking access via funcs. It will allow us in future to not care about Analytics init, we will track everything via same methods

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Tracking/AnalyticsManager.swift
mrkulik: @ValentinaIancu-Gini in that case I will be able to do it via generic, because we have 2 different enums for dict keys, not one `AnalyticsPropertyType`
Something like: `mapPropertiesToDictionary<T: RawRepresentable>(_ properties: [T: AnalyticsPropertyValue])`
Sounds good?

ValentinaIancu-Gini: if we want to include also
```
 for (property, value) in properties {
                superProperties[property] = value
            }
``` yes we need to use generics. 
I think is better like this than having duplicate-ish code.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Tracking/AnalyticsManager.swift
ValentinaIancu-Gini: @mrkulik true. I forgot the real reason of doing this. Thanks for reminding me this.

ValentinaIancu-Gini:
@mrkulik if you have issues with the merge conflicts we can stay together to fix them. Just let me know.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Tracking/AnalyticsManager.swift
mrkulik: @ValentinaIancu-Gini I created a reusable func and moved both suggested code blocks there. Thanks for your suggestions!

ValentinaIancu-Gini (approved):
Greate job @mrkulik !

PR #510, opened 2024-05-29
fix(GiniCaptureSDK): add camera back button event for custom bottom bar

mrkulik:
Add missing event to custom bottom bar


ValentinaIancu-Gini (approved):
@mrkulik Great job!

PR #519, opened 2024-06-10
Pp 518 add sdk opened and sdk closed events

ValentinaIancu-Gini:


BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
```diff
@@ -253,6 +254,11 @@ extension GiniBankNetworkingScreenApiCoordinator {
         }
     }
 
+    private func sendAnalyticsEventSDKClose() {
+        AnalyticsManager.track(event: .sdkClosed,
+                               properties: [AnalyticsProperty(key: .status, value: "successful")])
```
mrkulik, line 259: [Clarification] do we need to send somewhere event with `unsuccessful`?

mrkulik (commented):
@ValentinaIancu-Gini looks good👍
Thank you!

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
ValentinaIancu-Gini: @mrkulik for now no, but we want to easily add if we see that we need it.

PR #524, opened 2024-06-11
fix(GiniBankSDK): Remove duplicate `sdk_closed` event sent when tapping `Proceed` button in RA screen

ValentinaIancu-Gini:
…

PP-518


mrkulik (approved):

mrkulik (approved):
@ValentinaIancu-Gini will work for now, but after integrating with configuration service, we will be able to remove `firstSDKOpen`

ValentinaIancu-Gini:
@mrkulik I agree. we need to check together with the changes for the configuration service.

PR #525, opened 2024-06-11
Pp 537 disable analytics for users that denied the tracking permissions

ValentinaIancu-Gini:


zladzeyka (approved):
Looks good to me @ValentinaIancu-Gini 
Since we're adding `AdSupport` and `AppTrackingTransparency` I think we need to update the privacy manifest as well 🧐

ValentinaIancu-Gini:
> Looks good to me @ValentinaIancu-Gini Since we're adding `AdSupport` and `AppTrackingTransparency` I think we need to update the privacy manifest as well 🧐 And for making Sonar happy could you please add comment why the initializer is empty 😜 `private init() {} `

 `AdSupport` and `AppTrackingTransparency` are added in the SDK only to check the status of the permission. Anyway, I have a task to check if we need to update the privacy manifest. https://ginis.atlassian.net/browse/PP-504 
 
 I will look into Sonnar's suggestion. Thank you for the reminder!

mrkulik (approved):

PR #533, opened 2024-06-18
PP-539 Add all event to the queue

mrkulik:
PP-539

### DIFF
--- a/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Tracking/AnalyticsManager.swift
+++ b/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Tracking/AnalyticsManager.swift
@@ -95,30 +95,15 @@ public class AnalyticsManager {
     static func track(event: AnalyticsEvent,
                       screenNameString: String? = nil,
                       properties: [AnalyticsProperty] = []) {
-        guard let mixpanelInstance, amplitudeInitialised else {
-            let queuedEvent = QueuedAnalyticsEvent(event: event,
-                                                   screenNameString: screenNameString,
-                                                   properties: properties)
-            eventsQueue.append(queuedEvent)
-            return
+        let queuedEvent = QueuedAnalyticsEvent(event: event,
+                                               screenNameString: screenNameString,
+                                               properties: properties)
+        eventsQueue.append(queuedEvent)
+
+        // Process the event queue if SDKs are initialized
+        if mixpanelInstance != nil && amplitudeInitialised {
+            processEventsQueue()
         }
-
-        var eventProperties: [String: String] = [:]
-
-        if let screenName = screenNameString {
-            eventProperties[AnalyticsPropertyKey.screenName.rawValue] = screenName
-        }
-
-        for property in properties {
-            let propertyValue = property.value.analyticsPropertyValue()
-            eventProperties[property.key.rawValue] = convertPropertyValueToString(propertyValue)
-        }
-
-        // Track event in Mixpanel
-        mixpanelInstance.track(event: event.rawValue, properties: eventProperties)
-
-        // Track event in Ampltitude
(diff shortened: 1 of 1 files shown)


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Tracking/AnalyticsManager.swift
```diff
@@ -95,30 +95,15 @@ public class AnalyticsManager {
...
-            eventsQueue.append(queuedEvent)
-            return
+        let queuedEvent = QueuedAnalyticsEvent(event: event,
+                                               screenNameString: screenNameString,
+                                               properties: properties)
+        eventsQueue.append(queuedEvent)
+
+        // Process the event queue if SDKs are initialized
+        if mixpanelInstance != nil && amplitudeInitialised {
+            processEventsQueue()
```
ValentinaIancu-Gini, line 105: [clarification] Do we still to call this method in `    private static func handleAnalyticsSDKsInit() {` ? I think is not neccessary.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Tracking/AnalyticsManager.swift
```diff
@@ -136,9 +121,22 @@ public class AnalyticsManager {
     private static func processEventsQueue() {
```
ValentinaIancu-Gini, line 121: [suggestion] I think we should rename this method. What do you think about `trackEventsFromQueue` or `trackQueueEvents`?

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Tracking/AnalyticsManager.swift
```diff
@@ -136,9 +121,22 @@ public class AnalyticsManager {
...
+            for property in queuedEvent.properties {
+                let propertyValue = property.value.analyticsPropertyValue()
+                eventProperties[property.key.rawValue] = convertPropertyValueToString(propertyValue)
+            }
+
+            // Track event in Mixpanel
+            mixpanelInstance?.track(event: queuedEvent.event.rawValue, properties: eventProperties)
+
+            // Track event in Amplitude
+            ampltitudeTrackEvent(event: queuedEvent.event, eventProperties: eventProperties)
```
ValentinaIancu-Gini, line 139: [suggestion] Maybe we can put this portion of code into a separate method. What do you think?

mrkulik: @ValentinaIancu-Gini 
I think we can put this code in method track(event: QueuedAnalyticsEvent) and call it from `processEventsQueue`. In that way we will keep consistent structure of our methods
Works good?

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Tracking/AnalyticsManager.swift
mrkulik: @ValentinaIancu-Gini if we will move out tracking events code from this method and I suggest to choose something from this:
`releaseEventsQueue`
`processEventsQueue`
`flushEventsQueue`

ValentinaIancu-Gini: mhh let's keep it as it is`processEventsQueue` but maybe we can add a comment to this method and describe what is actually doing.

ValentinaIancu-Gini (approved):
@mrkulik good job.
Let's wait for the checks to pass and after that, we can merge this.
Thank you!

PR #537, opened 2024-06-20
PP 504 check if the privacy manifests needs to be updated

ValentinaIancu-Gini:
Update privacy manifest files with tracking information:
Apple documentation:

"At the top level of this property list file, add the following keys to the dictionary:

NSPrivacyTracking
A Boolean that indicates whether your app or third-party SDK uses data for tracking as defined under the App Tracking Transparency framework. For more information, see [User Privacy and Data Use](https://developer.apple.com/app-store/user-privacy-and-data-use/).

NSPrivacyTrackingDomains
An array of strings that lists the internet domains your app or third-party SDK connects to that engage in tracking. If the user has not granted tracking permission through the App Tracking Transparency framework, network requests to these domains fail and your app receives an error.

**To provide a list of internet domains in NSPrivacyTrackingDomains, set NSPrivacyTracking to true.** "

Check also the ticket comments https://ginis.atlassian.net/browse/PP-504  where you can find links to the Apple documentation.


mrkulik (approved):

mrkulik (approved):
Great job!

PR #539, opened 2024-06-20
fix(GinIHealthSDK): Handle error when loading PDF data IPC-280

razvancapra:
IPC-280
https://ginis.atlassian.net/browse/IPC-280


zladzeyka (approved):
Looks great,
thank you @razvancapra 👍

PR #541, opened 2024-06-20
fix(GiniBankSDK): Sent `help_tapped` event from RA screen when `custom navigation bottom bar` is enabled

ValentinaIancu-Gini:
Sent `help_tapped` event from RA screen when `custom navigation bottom bar` is enabled
PP-582


mrkulik (approved):
@ValentinaIancu-Gini 🙌

PR #542, opened 2024-06-21
PP 585 UJ return assistant enabled user property isn't show value from configuration api request

ValentinaIancu-Gini:


mrkulik (approved):
@ValentinaIancu-Gini thank you for fixing it!👍

ValentinaIancu-Gini:
> @ValentinaIancu-Gini thank you for fixing it!👍

Thank you, I would merge this after Vlad is testing it and everything is ok.
This should be our process from now on if possible.

PR #546, opened 2024-06-24
Merchant sdk refactor credentials

motokotomax:
Proposition to refactor: move credentials from plist to code. 
+ less files and code
+ file with credentials will be ignore by git 
+ developer will be notified with crash, if he don't fill own creds
- need to update documentation, remove image with creds format


razvancapra (approved):
👍 Nice refactor. If that's the way we do want to provide credentials. If this is the new way to offer the authentication option, I'm ok with that

zladzeyka (approved):
Looks good to me!

motokotomax:
Thanks for review, but initialization method changed, and this refactoring not needed any more

PR #553, opened 2024-06-28
Ipc 291 fix share with flow

a-szotyori:
Moves the publishing of the paymentRequestID in the "share with" case to when the user has tapped on an app in the native share sheet for the "share with" PDF. This way when clients can dismiss the Payment Review Screen on receiving the paymentRequestID safely also for the "share with" case.

The [bug ticket](https://ginis.atlassian.net/browse/IPC-289) contains more details.

### DIFF
--- a/HealthSDK/GiniHealthSDKExample/GiniHealthSDKExample/AppCoordinator.swift
+++ b/HealthSDK/GiniHealthSDKExample/GiniHealthSDKExample/AppCoordinator.swift
@@ -330,6 +330,10 @@ extension AppCoordinator: GiniHealthDelegate {
     
     func didCreatePaymentRequest(paymentRequestID: String) {
         print("✅ Created payment request with id \(paymentRequestID)")
+        DispatchQueue.main.async {
+            (self.childCoordinators.first as! InvoicesListCoordinator).invoicesListViewController
+                .presentedViewController?.dismiss(animated: true)
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


HealthSDK/GiniHealthSDKExample/GiniHealthSDKExample/AppCoordinator.swift
```diff
@@ -330,6 +330,10 @@ extension AppCoordinator: GiniHealthDelegate {
     
     func didCreatePaymentRequest(paymentRequestID: String) {
         print("✅ Created payment request with id \(paymentRequestID)")
+        DispatchQueue.main.async {
+            (self.childCoordinators.first as! InvoicesListCoordinator).invoicesListViewController
```
razvancapra, line 334: Recommendation: To avoid using forced unwrap, I suggest using this approach:

`guard let invoicesListCoordinator = self.childCoordinators.first as? InvoicesListCoordinator else {
                return
            }
            invoicesListCoordinator.invoicesListViewController.presentedViewController?.dismiss(animated: true)`

razvancapra (changes requested):
Other than that. Nice work 👍

zladzeyka (approved):
👍🏻

caprapaditu (approved):

razvancapra (approved):

PR #568, opened 2024-07-05
Pp 631 add skonto screen in the sdk flow

ValentinaIancu-Gini:


BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/Skonto/Views/SkontoProceedView.swift
```diff
@@ -8,7 +8,7 @@ import UIKit
 import GiniCaptureSDK
 
 class SkontoProceedView: UIView {
-    private lazy var payButton: MultilineTitleButton = {
+    lazy var proceedButton: MultilineTitleButton = {
```
mrkulik, line 11: [Suggestion] lets keep it private and add action to init, something like:

```
private var proceedAction: (() -> Void)?`

    init(proceedAction: (() -> Void)?) {
        self.proceedAction = proceedAction
    }
```

The main goal not to add target on public button out of view, we will have target setup in view and will call injected action

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/Skonto/SkontoViewController.swift
```diff
@@ -125,6 +131,7 @@ public class SkontoViewController: UIViewController {
         notAppliedGroupView.addSubview(notAppliedView)
         view.addSubview(proceedView)
 
+        proceedView.proceedButton.addTarget(self, action: #selector(proceedButtonTapped), for: .touchUpInside)
```
mrkulik, line 134: [Suggestion] lets keep addTarget inside view and just inject action in view init (example will be in a comment below)

mrkulik (changes requested):
@ValentinaIancu-Gini looks good, but seems like we can finish it only after backend integration. For current moment just 1 change with actions inject
Nice job!

mrkulik (approved):

PR #569, opened 2024-07-05
fix(GiniMerchantSDK): Update Xcode for check workflow

zladzeyka:
EC-63
 - Update Xcode for the check workflow
 - Add params to the pinning example scheme


motokotomax:
Awesome! thank you!

motokotomax (approved):

PR #576, opened 2024-07-11
fix(GiniBankSDKExample): Add PhotosUI framework

motokotomax:
Add PhotosUI framework to fix crash with limited photos access 
PP-510 
PP-519

Need to check and update documentation

reproducable only on device. use steps in PP-519


motokotomax:
GiniCaptureSDK use in files GalleryCoordinator.swift, AlbumsPickerViewController.swift
import Photos
import PhotosUI

Can’t find docs about PhotosUI, but seems to it contain UI for limited access gallery and UI libs don’t links automatically 

Checked on iPhone X (not reproducalbe on simulator), with import app don’t crash.

as SwiftPackage don't provide ability to link system frameworks (no project file) - only way is to import framewors PhotosUI and Photos to example/client app

zladzeyka (approved):
👍🏻 Thank you @motokotomax

PR #589, opened 2024-07-24
PP-630 update localization strings

mrkulik:
PP-630


ValentinaIancu-Gini (commented):
@mrkulik everything looks good!

I think we should wait a bit until we can merge this since we are still waiting for some confirmations of some texts.

PR #591, opened 2024-07-24
Pp 646 add bank sdk version and capture sdk version as user properties

ValentinaIancu-Gini:
We decide to not set anymore the app_version and add bank_sdk_version and capture_sdk_version as user properties.


mrkulik (commented):
@ValentinaIancu-Gini perfect!

PR #592, opened 2024-07-25
PP 602 Skonto iOS

ValentinaIancu-Gini:
Development of the MVP for the Skonto feature.
More details about this can be found in the epic: https://ginis.atlassian.net/browse/PP-602.


ValentinaIancu-Gini:
I've checked sonarcloud suggestions and besides some code comment out because we decided to not deliver help functionality in this MVP and one TODO related to the color scheme that we want to introduce we have one small thing to fix: `// TODO: To specify what we exactly need to expose`

mrkulik (approved):

PR #596, opened 2024-07-25
feat(GiniMerchantSDK): Payment component 2 rows

razvancapra:
EC-92

+ Configurable so that we can still use 1 row when needed
+ Ability to play with settings in DebugViewController: Show Payment component in 1 row & Hide information if user used paayment component before

Task included: https://ginis.atlassian.net/browse/EC-92

### DIFF
--- a/MerchantSDK/GiniMerchantSDK/Sources/GiniMerchantSDK/Core/PaymentComponent/BanksBottomView.swift
+++ b/MerchantSDK/GiniMerchantSDK/Sources/GiniMerchantSDK/Core/PaymentComponent/BanksBottomView.swift
@@ -172,7 +172,8 @@ class BanksBottomView: BottomSheetViewController {
             bottomStackView.leadingAnchor.constraint(equalTo: bottomView.leadingAnchor, constant: Constants.viewPaddingConstraint),
             bottomStackView.trailingAnchor.constraint(equalTo: bottomView.trailingAnchor, constant: -Constants.viewPaddingConstraint),
             bottomStackView.topAnchor.constraint(equalTo: bottomView.topAnchor, constant: Constants.topAnchorPoweredByGiniConstraint),
-            bottomStackView.bottomAnchor.constraint(equalTo: bottomView.bottomAnchor)
+            bottomStackView.bottomAnchor.constraint(equalTo: bottomView.bottomAnchor),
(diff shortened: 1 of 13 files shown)


MerchantSDK/GiniMerchantSDK/Sources/GiniMerchantSDK/Core/PaymentComponent/BanksBottomView.swift
```diff
@@ -191,6 +192,7 @@ extension BanksBottomView {
         static let closeIconSize = 24.0
         static let titleViewTitleIconSpacing = 10.0
         static let topAnchorPoweredByGiniConstraint = 5.0
+        static let bottomViewHeight = 44.0
```
motokotomax, line 195: Maybe put this constant into common config or constants for all PaymentComponents?

MerchantSDK/GiniMerchantSDK/Sources/GiniMerchantSDK/Core/PaymentComponent/PaymentComponentConfiguration.swift
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     Set to `true` to make see payment component in 1 row instead of 2
+     */
+    public var showPaymentComponentInOneRow = false
+
+    /**
+     Set to `true` to hide information like select your bank title label and more information view if user is returning and used component multiple times
+     */
+    public var hideInfoForReturningUser = false
```
motokotomax, line 24: we pass default values in init - so this vars can be just :Bool (without default values)

motokotomax (approved):
Thanks @razvancapra, please have a look at my optional suggestions

MerchantSDK/GiniMerchantSDK/Sources/GiniMerchantSDK/Core/PaymentComponent/BanksBottomView.swift
caprapaditu: Good idea, but for now we don't have a common config. All constants are private in their views. For this one just happened to be the same in all places, but they might change

zladzeyka (changes requested):
Thank you @razvancapra 
looks good!
Please find some small notes from my side:

zladzeyka (approved):
Looks good @razvancapra 👏🏻

PR #600, opened 2024-07-26
Add Utils package

motokotomax:


ValentinaIancu-Gini:
@motokotomax I have one question...when we need to update this Uitls package, we will need to version it, right? Is not adding complexity to what we already have? Or can somehow skip the versioning part of that specific package?

motokotomax:
> @motokotomax I have one question...when we need to update this Uitls package, we will need to version it, right? Is not adding complexity to what we already have? Or can somehow skip the versioning part of that specific package?

I think we can use just main branch (without version). but seems to code in utils package is "pretty constant"
Screenshot 2024-07-29 at 10 51 59

GiniComponents/GiniUtilites/Sources/GiniUtilites/Utilites.swift
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ Returns a decimal value
+
+ - parameter inputFieldString: String from input field.
+
+ - returns: decimal value in current locale.
+ */
+
+public func decimal(from inputFieldString: String) -> Decimal? {
```
razvancapra, line 17: Suggestion: Maybe we should add this to a String extension into our SDK?!

razvancapra (approved):
Nice job, Max! 👍

GiniComponents/GiniUtilites/Sources/GiniUtilites/Utilites.swift
motokotomax: done

PR #607, opened 2024-07-31
PP-639: Attempt to fix the UIDocument crash

igor-gini:
I still didn't manage to catch the crash, but judging by the error from the video, I think we should just overwrite the writeContents method and make a check, similar to what is done in `loadFromContents` in `InputDocument` (as far as I understood it is used instead of the regular `UIDocument`).

### DIFF
--- a/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniCaptureDocument.swift
+++ b/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniCaptureDocument.swift
@@ -112,13 +112,20 @@ private extension GiniCaptureDocumentBuilder {
     final class InputDocument: UIDocument {
         public var data: Data?
 
-        override public func load(fromContents contents: Any, ofType typeName: String?) throws {
+        override func load(fromContents contents: Any, ofType typeName: String?) throws {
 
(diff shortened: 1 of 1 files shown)


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniCaptureDocument.swift
```diff
@@ -120,5 +120,12 @@ private extension GiniCaptureDocumentBuilder {
 
             self.data = data
         }
+
```
zladzeyka, line 123: [Clarification] @igor-gini could you please check if we still need `public` in `override public func load()`

zladzeyka (commented):
Thank you @igor-gini and congratulations on the 1st PR here 😊
Could you please 1 small note below?

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniCaptureDocument.swift
igor-gini: I don't think we need it to be public, removed it there

zladzeyka (approved):
👍🏻

PR #610, opened 2024-08-01
fix(GiniBankSDK): add paste-only mode for price textField. Paste vali…

mrkulik:
…dation

PP-645


ValentinaIancu-Gini:
@mrkulik Something is not working for the first time when I try to paste in the "With Skonto" textfield. See the video attached.


Uploading RPReplay_Final1722589237.MP4…

mrkulik:
> @mrkulik Something is not working for the first time when I try to paste a text into the "With Skonto" text field. See the video attached.
> 
> https://drive.google.com/file/d/1N15ROf25mjQtsOw0159cgIY5oftTUbWu/view?usp=drive_link

Lets handle it in terms of validation message task. For now, Skonto text field automatically declines copy/paste if value is more than original one

PR #612, opened 2024-08-02
PP-523 fix camera issue when going back from help screen

igor-gini:
Prevent reinitializing camera every time it gets opened until it gets deallocated


zladzeyka:
@igor-gini that you for the PR, I've started with tests and discovered a scenario where the camera is not initialized
https://github.com/user-attachments/assets/4edd72ae-32ab-41fc-b302-3975565c1f55

igor-gini:
@zladzeyka I've addressed it in another fix today

zladzeyka (approved):
Looks good! 👍🏻
We can move it to QA

zladzeyka:
@igor-gini could you please run the unit tests locally
Since you've extended `protocol CameraProtocol: AnyObject` you need to update the tests.
 scheme is `GiniCaptureSDKTests`

PR #618, opened 2024-08-09
PP-711 add clientid and clientsecret credentials

igor-gini:


zladzeyka (changes requested):
Thank you @igor-gini 
IMG_315468C3AD82-1

Could you please take about dark mode and make password as a isSecureTextEntry?

PR #624, opened 2024-08-16
EC-112 Layout fixes for Payment Review Screen

antosj:
EC-112


MerchantSDK/GiniMerchantSDK/Sources/GiniMerchantSDK/Core/PaymentComponent/PaymentComponentsController.swift
```diff
@@ -49,6 +50,7 @@ public final class PaymentComponentsController: PaymentComponentsProtocol {
 
     /// Payment Component View Configuration
     public var paymentComponentConfiguration: PaymentComponentConfiguration?
+    public var paymentReviewViewController: PaymentReviewViewController?
```
razvancapra, line 53: Suggestion: I won't store this viewController into the controller, I would always instantiate and pass it to the user.

MerchantSDK/GiniMerchantSDK/Sources/GiniMerchantSDK/Core/PaymentComponent/PaymentComponentsController.swift
```diff
@@ -191,17 +193,22 @@ public final class PaymentComponentsController: PaymentComponentsProtocol {
...
+                                                                              paymentInfo: paymentInfo,
+                                                                              selectedPaymentProvider: selectedPaymentProvider,
+                                                                              trackingDelegate: trackingDelegate,
+                                                                              paymentComponentsController: self)
+        completion(paymentReviewViewController, nil)
     }
 
     // MARK: - Bottom Sheets
 
+    public func paymentInfoBottomSheet() -> UIViewController {
```
razvancapra, line 208: Suggestion: I would keep using: `loadPaymentReviewScreenWithoutDocument` and/or if mandatory, I would rename this function to something like: `paymentReviewBottomSheet`.

We already have `paymentInfoViewController`, which is about something else. It might be misleading.

MerchantSDK/GiniMerchantSDK/Sources/GiniMerchantSDK/Core/PaymentReviewModel.swift
```diff
@@ -40,48 +37,22 @@ public class PaymentReviewModel: NSObject {
...
-    
-    var isImagesLoading: Bool = false {
-        didSet {
-            self.updateImagesLoadingStatus?()
-        }
-    }
 
     var paymentComponentsController: PaymentComponentsController
 
-    public init(with giniMerchant: GiniMerchant, document: Document?, extractions: [Extraction]?, paymentInfo: PaymentInfo?, selectedPaymentProvider: PaymentProvider?, paymentComponentsController: PaymentComponentsController) {
```
razvancapra, line 67: Clarification: You removed the PaymentReview with document logic, but we will merge both implementations in the next phase. I think we are supposed to keep them both and use them based on a small configuration, or on the SDK

MerchantSDK/GiniMerchantSDK/Sources/GiniMerchantSDK/Core/PaymentReviewViewController+UICollection.swift
```diff
@@ -1,42 +0,0 @@
...
-//
-//  Copyright © 2024 Gini GmbH. All rights reserved.
-//
-
-import UIKit
-import GiniUtilites
-
-// MARK: - UICollectionViewDelegate, UICollectionViewDataSource
-
-extension PaymentReviewViewController: UICollectionViewDelegate, UICollectionViewDataSource, UICollectionViewDelegateFlowLayout {
```
razvancapra, line 13: Clarification: We will still need these in next phase when all these will go in Payment Component SDK that will be used commonly by GiniHealthSDK and GiniMerchantSDK

razvancapra (changes requested):
Also, `minHeight` for bottom sheets doesn't work anymore as was supposed to work (covering the previous bottom sheet)

razvancapra:
@antosj also UIActivityViewController for sharing the PDF flow is not working anymore: 
```Attempt to present <UIActivityViewController: 0x10c809800> on <GiniMerchantSDK.PaymentReviewViewController: 0x10f008200> (from <GiniMerchantSDK.PaymentReviewViewController: 0x10f008200>) whose view is not in the window hierarchy.```

PR #625, opened 2024-08-19
fix(GiniBankSDK): fix font size for help item

mrkulik:
PP-684


ValentinaIancu-Gini (approved):
@mrkulik Looks good! Thank you!

PR #633, opened 2024-08-22
fix(GiniBankSDK): Fix the display of the Skonto edge case alert info view when tapping on the info banner view

ValentinaIancu-Gini:
We introduced this bug when we added `hasShownAlert` to not show once again the alert for Skonto edge cases when coming back from Help.
PP-738


mrkulik (commented):
@ValentinaIancu-Gini 👍

mrkulik (approved):

PR #638, opened 2024-08-26
fix(GiniBankSDK): Fix image overlap when uploading screenshots between document pictures

ValentinaIancu-Gini:
PP-757


mrkulik (approved):
@ValentinaIancu-Gini 👍

PR #639, opened 2024-08-27
PP-160 Fix border color not updating dynamically

igor-gini:


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/ButtonConfiguration.swift
```diff
@@ -61,11 +61,22 @@ extension BottomLabelButton {
...
+
+public extension GiniCaptureButton {
     func configure(with configuration: ButtonConfiguration) {
         self.backgroundColor = configuration.backgroundColor
         self.layer.borderColor = configuration.borderColor.cgColor
         self.layer.shadowColor = configuration.shadowColor.cgColor
+        traitCollectionUpdated = { [weak self] in
+            self?.layer.borderColor = configuration.borderColor.cgColor
+            self?.layer.shadowColor = configuration.shadowColor.cgColor
+        }
```
zladzeyka, line 79: [Clarification]: @igor-gini why do we update only `self?.layer.borderColor` and `self?.layer.shadowColor`? Is it because we need a special update for self.layer?

zladzeyka (commented):
Thank you very much @igor-gini 
One tiny clarifiying question from my side below.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/ButtonConfiguration.swift
igor-gini: This is because colors in CALayer are CGColors, which can’t dynamically adapt to the current trait collection like UIColor does, so we need to update them each time the trait collection updates to ensure they are always correct

zladzeyka (approved):
Looks good!
Please prepare a build for QA

PR #644, opened 2024-08-28
PP-161 fix separator in the last cell

igor-gini:


zladzeyka (changes requested):
I've tried to test the branch and I've got the weird results:
IMG_7542
IMG_7545

IMG_7544

IMG_7543

I've notices in the debug view message:

`Changing the background color of UITableViewHeaderFooterView is not supported. Use the background view configuration instead.`
 @igor-gini could you please take care about it?

igor-gini:
Fixed the separators issue, seems to be all good now ‌

Regarding that “changing background color is unsupported“ issue: it seems that it happens with AlbumsHeaderView (the warning seems to disappear when I comment it out), but I don’t see it setting the background color anywhere, the only place where background color is set there are the viewcontroller’s view background color and table view’s background color, which don’t seem to be causing the issue. I think it could be the NIB file setting it and causing the warning to appear, since AlbumsHeaderView gets loaded from NIB

zladzeyka (commented):
When you have a couple of albums and add one more 
The delimiter between 1 and 2 will disappear
IMG_7577

igor-gini:
> When you have a couple of albums and add one more
> The delimiter between 1 and 2 will disappear

It's likely because the default separator inset isn’t set anywhere, and after it’s set to be not visible, nothing sets it back. Added setting default inset in setUp of the AlbumsPickerTableViewCell

zladzeyka (approved):
Thank you @igor-gini 
Let's prepare a build for QA

zladzeyka (approved):
Thank you

Let's prepare a build for QA

PR #655, opened 2024-09-04
PP-401 fix camera getting stuck on share when action sheet is presented

igor-gini:


zladzeyka (approved):
A duplication for PR shall we keep this one and close a previous one?

PR #656, opened 2024-09-04
Pp 719 create a flag in sdk to enable disable the transaction docs feature

ValentinaIancu-Gini:


mrkulik (approved):
@ValentinaIancu-Gini 👍

PR #640, opened 2024-08-27
PP-166 fix digital invoice line item wrong background color

igor-gini:

PR #658, opened 2024-09-05
PP-802: Fix crash in storyboard

igor-gini:
The reason was that in the storyboard of that button it didn't have any custom class, just UIButton, although in the code was MultilineTitleButton, it did not crash simply because there were no additional properties before adding GiniCaptureButton.

EXC_BAD_ACCESS with code 1 usually means that the object either does not exist in memory at all (either it was never allocated or it was freed), or the address of the memory location is crooked. In this case, because the class was not defined in the storyboard (so inside uikit the memory was allocated just for UIButton), the memory for traitCollectionUpdated was not allocated, even if you put nil as default value (and Optional != nullptr).

Sometimes EXC_BAD_ACCESS also occurs because of some bugs in xcode (so far I've seen it only with v15.3.0), cleaning build folder/derived data helps in this case.


zladzeyka (changes requested):
There is one more place to fix
Screenshot 2024-09-09 at 15 50 39

igor-gini:
> There is one more place to fix Screenshot 2024-09-09 at 15 50 39

This has been fixed and merged in [PP-166](https://github.com/gini/gini-mobile-ios/pull/640), I've rebased on the latest main, so this PR should have both fixes in

ValentinaIancu-Gini (approved):
@igor-gini I tested. All is good. Thank you!

zladzeyka (approved):
Thank you @igor-gini

PR #670, opened 2024-09-24
Pp 830 i os td UI adjustment to allow customers more flexibility e.g. remove borders

mrkulik:


ValentinaIancu-Gini (approved):
@mrkulik Nice work. Thank you!

PR #671, opened 2024-09-25
Pp 809 see the attached invoices on the transaction summary screen

ValentinaIancu-Gini:


BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
```diff
@@ -532,6 +555,27 @@ extension GiniBankNetworkingScreenApiCoordinator: SkontoCoordinatorDelegate {
...
+            self?.transactionDocsDataCoordinator?.transactionDocs = []
+            deliveryFunction(extractionResult)
+        },
+                                                attachAction: { [weak self] in
+            if let documentId = documentId {
+                self?.transactionDocsDataCoordinator?.transactionDocs = [.init(documentId: documentId,
+                                                                               fileName: "Document",
+                                                                               type: .document)]
+            } else {
+                self?.transactionDocsDataCoordinator?.transactionDocs = []
```
ValentinaIancu-Gini, line 573: [clarification] @mrkulik do we really need to set to an empty array?

mrkulik: For don't attach yes, view is set up to be hidden if there are empty docs
Here it's to handle a case when we don't have documentId. I'm not sure we can face it really, but to be able to handle this case I added this option as default value

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
```diff
@@ -233,6 +241,13 @@ open class GiniBankNetworkingScreenApiCoordinator: GiniScreenAPICoordinator, Gin
         })
         return self.start(withDocuments: documents, animated: animated)
     }
+
+    /// Attempts to cast the public-facing `transactionDocsDataCoordinator` to the internal protocol.
+    /// - Returns: The `TransactionDocsDataInternalProtocol` instance if the cast is successful, otherwise `nil`.
+    private func internalTransactionDocsDataCoordinator() -> TransactionDocsDataInternalProtocol? {
```
mrkulik, line 247: [Suggestion] this method is not creating a new entity, it returns existing one with typecast. I suggest here to use computed property and remove it from inits of GiniBankNetworkingScreenApiCoordinator

```
/// Attempts to cast the public-facing `transactionDocsDataCoordinator` to the internal protocol.
/// - Returns: The `TransactionDocsDataInternalProtocol` instance if the cast is successful, otherwise `nil`.
private var internalTransactionDocsDataCoordinator: TransactionDocsDataInternalProtocol? {
    return GiniBankConfiguration.shared.transactionDocsDataCoordinator as? TransactionDocsDataInternalProtocol
}

```

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/TransactionDocs/TransactionDocsDataCoordinator.swift
```diff
@@ -0,0 +1,131 @@
...
+    /// - Returns: A `Bool` representing whether documents should always be attached to the transaction.
+    func getAlwaysAttachDocsValue() -> Bool
+
+    /// Sets the "Always Attach Documents" setting to a given value.
+    /// - Parameter value: A `Bool` indicating whether documents should always be attached to the transaction.
+    func setAlwaysAttachDocs(_ value: Bool)
+
+    /// Checks if there are any attached transaction documents.
+    /// - Returns: A `Bool` indicating whether there are any attached documents.
+    func hasAttachedDocuments() -> Bool
```
mrkulik, line 27: [Clarification] should we replace it with transactionDocsIds: [String]? If we need to provide ids of attached documents

mrkulik (changes requested):
@ValentinaIancu-Gini thank you! I left some suggestions, can you please check them

mrkulik (approved):
@ValentinaIancu-Gini 👍

PR #674, opened 2024-09-25
feat(GiniHealthSDK): Expose doctor's name 

razvancapra:
IPC-332

### DIFF
--- a/HealthAPILibrary/GiniHealthAPILibrary/Sources/GiniHealthAPILibrary/Documents/ExtractionResult.swift
+++ b/HealthAPILibrary/GiniHealthAPILibrary/Sources/GiniHealthAPILibrary/Documents/ExtractionResult.swift
@@ -26,6 +26,8 @@ public enum ExtractionType: String {
     case iban = "iban"
     case paymentPurpose = "payment_purpose"
     case doctorName = "medical_service_provider"
+    case bic = "bic"
+    case invoiceDate = "invoice_date"
 }
 
 /**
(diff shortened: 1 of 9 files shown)


HealthAPILibrary/GiniHealthAPILibrary/Sources/GiniHealthAPILibrary/Documents/ExtractionResult.swift
```diff
@@ -25,6 +25,7 @@ public enum ExtractionType: String {
     case paymentRecipient = "payment_recipient"
     case iban = "iban"
     case paymentPurpose = "payment_purpose"
+    case doctorName = "medical_service_provider"
```
zladzeyka, line 28: [Clarification]: Do we use it in SDK? As I understood no, means we have here a convenience enum, is it correct?

zladzeyka (commented):
Awesome @razvancapra 
Please find my clarification comment

HealthAPILibrary/GiniHealthAPILibrary/Sources/GiniHealthAPILibrary/Documents/ExtractionResult.swift
razvancapra: Yes, it's a helper used for clients now.  Because we still haven't migrated to BaseSDKs. We use this in GiniHealthSDKExample app as an example

zladzeyka: https://health-api.gini.net/documentation/v4/#extractions
could you please add to the helper 2 more cases for `bic` and `invoice_date` then we will have all keys in 1 place

zladzeyka (changes requested):
There's 1 tiny improvement before we move it to QA:

zladzeyka (approved):
@razvancapra  thank you
Let's add a build for QA

PR #682, opened 2024-09-30
fix(GiniHealthSDKExample): Fix To the banking app button should be disabled

razvancapra:
IPC-338


zladzeyka (approved):
Thanks @razvancapra, please prepare a build for QA

PR #684, opened 2024-10-03
IPC-322 HealthAPILibrary and HealthSDK XCFrameworks

antosj:


HealthSDK/GiniHealthSDK/Sources/GiniHealthSDK/Core/GiniHealthUtils.swift
```diff
@@ -11,7 +11,7 @@ import UIKit
  
  */
 public func giniHealthBundle() -> Bundle {
-    Bundle.module
+    Bundle.resource
```
razvancapra, line 14: Clarification: Why we need this change? 
We have down here the `giniHealthBundleResource()` function that's providing the resources.

razvancapra (changes requested):
Look good. 
Just a small clarification.

HealthSDK/GiniHealthSDK/Sources/GiniHealthSDK/Core/GiniHealthUtils.swift
```diff
@@ -11,7 +11,7 @@ import UIKit
  
  */
 public func giniHealthBundle() -> Bundle {
-    Bundle.module
+    giniHealthBundleResource()
```
caprapaditu, line 14: Regarding: `giniHealthBundle()`
I would rather leave that as it is and I would use `giniHealthBundleResource()` in the `defaultImage()` function. I see there is the only place where is used.

caprapaditu (changes requested):

HealthSDK/GiniHealthSDK/Sources/GiniHealthSDK/Core/GiniHealthUtils.swift
razvancapra, line 14: Regarding: `giniHealthBundle()`
I would rather leave that as it is and I would use `giniHealthBundleResource()` in the `defaultImage()` function. I see there is the only place where is used.

razvancapra (changes requested):

razvancapra (approved):

PR #691, opened 2024-10-11
Pp 656 add integration tests for skonto skonto and ra

ValentinaIancu-Gini:


mrkulik (approved):
@ValentinaIancu-Gini looks good!
Just want to suggest to split files into folders in Tests, like you've done for Resources

mrkulik (approved):
@ValentinaIancu-Gini Thank you!👍

PR #696, opened 2024-10-21
PP-774 fix process button getting active too early when multiple documents

igor-gini:


zladzeyka (approved):
Looks good to me
Thanks @igor-gini

ValentinaIancu-Gini (approved):
@igor-gini Looks good to me as well! Thank you!

PR #697, opened 2024-10-21
ci: Change push triggers to pull_request

GeraltRiv:
PP-850

### DIFF
--- a/.github/workflows/capture-sdk.build.docs.yml
+++ b/.github/workflows/capture-sdk.build.docs.yml
@@ -1,13 +1,11 @@
 name: Builds docs for Capture SDK
 
 on:
-  push:
-    paths:
-      - 'CaptureSDK/**'
   workflow_dispatch:
     tags-ignore:
       - '**'
   pull_request:
+    types: [opened, edited, reopened, synchronize]
     paths:
       - 'CaptureSDK/**'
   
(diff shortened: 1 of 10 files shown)


zladzeyka (approved):
Thank you @GeraltRiv

zladzeyka:
@ValentinaIancu-Gini  fyi 😇

.github/workflows/capture-sdk.build.docs.yml
```diff
@@ -1,13 +1,11 @@
 name: Builds docs for Capture SDK
 
 on:
-  push:
-    paths:
-      - 'CaptureSDK/**'
   workflow_dispatch:
     tags-ignore:
```
ValentinaIancu-Gini, line 5: [clarification]Why for most of the checks we removed this and for this not? Mostly `docs check` has this.

ValentinaIancu-Gini (commented):
@GeraltRiv awesome! Thank you!
I left a small clarification question.

.github/workflows/capture-sdk.build.docs.yml
GeraltRiv: Because it's just a build.docs. It doesn't need to run on every push. 
But publish docs still have push trigger

ValentinaIancu-Gini: Thank you for the clarification!

PR #698, opened 2024-10-21
ci: Add concurrency

GeraltRiv:
PP-851

### DIFF
--- a/.github/workflows/bank-sdk.publish.docs.yml
+++ b/.github/workflows/bank-sdk.publish.docs.yml
@@ -11,10 +11,6 @@ on:
         required: true
   workflow_dispatch:
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: ${{ !contains(github.ref, 'refs/tags/')}}
-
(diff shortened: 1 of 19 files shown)


.github/workflows/bank-sdk.publish.docs.yml
```diff
@@ -11,6 +11,10 @@ on:
         required: true
   workflow_dispatch:
 
+concurrency:
```
zladzeyka, line 14: [Suggestion]: @GeraltRiv it's a part of release workflow, let's skip concurrency for `*.publish.docs.yml`  files

zladzeyka (changes requested):
@GeraltRiv Thank you very much 🎉
I have one suggestion for excluding the concurrency from publishing docs below:

.github/workflows/bank-sdk.publish.docs.yml
GeraltRiv: Fixed

GeraltRiv: Fixed

PR #711, opened 2024-10-30
Pp 680 i os skonto screen add input fields validation

mrkulik:


BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/Skonto/Components/PriceTextField.swift
```diff
@@ -56,8 +57,8 @@ class PriceTextField: UITextField, UITextFieldDelegate {
         return false
     }
 
-    private func filterAndTrimInput(_ text: String) -> String {
-        return String(text.trimmingCharacters(in: .whitespaces).filter { $0.isNumber }.prefix(7))
+    private func filterInput(_ text: String) -> String {
```
ValentinaIancu-Gini, line 60: [suggestion]: Can we rename it like: `filterTextInput`?

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/Skonto/Models/SkontoViewModel.swift
```diff
@@ -151,23 +158,52 @@ class SkontoViewModel {
     }
 
     func setSkontoAmountToPayPrice(_ price: String) {
-        guard let price = convertPriceStringToPrice(price: price),
-              price.value <= amountToPay.value else {
+        setPrice(
+            price,
+            maxValue: amountToPay.value,
+            errorMessage: NSLocalizedStringPreferredGiniBankFormat("ginibank.skonto.withdiscount.validation",
```
ValentinaIancu-Gini, line 164: [suggestion] I think for better readability is better to have a separate let for the `errorMessage` value and to fix the indentation for the `setPrice` method. What do you say?

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/Skonto/Models/SkontoViewModel.swift
```diff
@@ -151,23 +158,52 @@ class SkontoViewModel {
...
+            updateDocumentPagesModelData()
+            recalculateSkontoPercentage()
+        }
+    }
+
+    func setAmountToPayPrice(_ price: String) {
+        setPrice(
+            price,
+            maxValue: maximumAmountToPayValue,
+            errorMessage: NSLocalizedStringPreferredGiniBankFormat("ginibank.skonto.withoutdiscount.validation",
```
ValentinaIancu-Gini, line 177: [suggestion] I think for better readability is better to have a separate let for the `errorMessage` value and to fix the indentation for the `setPrice` method. What do you say?

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/Skonto/Models/SkontoViewModel.swift
```diff
@@ -151,23 +158,52 @@ class SkontoViewModel {
...
+                                                                   comment: "Your transfer limit has been exceeded: %@")
+        ) { validatedPrice in
+            amountToPay = validatedPrice
+            recalculateAmountToPayWithSkonto()
+            updateDocumentPagesModelData()
+        }
+    }
+
+    private func setPrice(_ price: String, maxValue: Decimal, errorMessage: String, completion: (Price) -> Void) {
```
ValentinaIancu-Gini, line 186: [request] Please move each parameter (except the first one) to a new line.

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/Skonto/Models/SkontoViewModel.swift
```diff
@@ -124,11 +126,16 @@ class SkontoViewModel {
...
     }
 
+    private var errorMessage: String?
+
     weak var delegate: SkontoViewModelDelegate?
 
-    init(skontoDiscounts: SkontoDiscounts, isWithDiscountSwitchAvailable: Bool = true) {
+    init(skontoDiscounts: SkontoDiscounts,
+         isWithDiscountSwitchAvailable: Bool = true,
+         maximumAmountToPayValue: Decimal = 99999.99) {
```
ValentinaIancu-Gini, line 135: [suggestion] Can we have the maximumAmountToPayValue of 99999.99 as a let property?

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/Skonto/Views/SkontoAmountToPayView.swift
```diff
@@ -132,6 +171,19 @@ class SkontoAmountToPayView: UIView {
...
+    func updateValidationMessage(_ message: String) {
+        guard validationLabel.isHidden else {
+            return
+        }
+        validationLabel.text = message
+        validationLabel.isHidden = false
+    }
+
+    func hideValidationMessage() {
+        validationLabel.text = ""
```
ValentinaIancu-Gini, line 183: [clarification] Why setting to nil is not good here instead of empty string?

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/Skonto/Views/SkontoWithDiscountPriceView.swift
```diff
@@ -25,12 +25,17 @@ class SkontoWithDiscountPriceView: UIView {
...
         bindViewModel()
+        setupKeyboardObservers()
     }
 
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
 
+    deinit {
+        NotificationCenter.default.removeObserver(self)
```
ValentinaIancu-Gini, line 36: [suggestion] Maybe is better to remove that specific observer that we added like:
`  NotificationCenter.default.removeObserver(self, name: UIResponder.keyboardWillHideNotification, object: nil)` what do you say?

ValentinaIancu-Gini (commented):
@mrkulik Thank you for this pr.
I left some suggestions and questions. Can you please check them?

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/Skonto/Models/SkontoViewModel.swift
mrkulik: in the PR with PP-906 I have implemented `func setMaximumAmountToPayValue(_ value: Decimal?)` which changes this value for Skonto + RA, thats the reason of `var` here

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/Skonto/Views/SkontoAmountToPayView.swift
mrkulik: here's no matter `nil` or "", because we are not using anywhere check for nil or empty string. Its just to be sure that previous value is not showed to a user

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/Skonto/Views/SkontoWithDiscountPriceView.swift
mrkulik: I think on deinit it's better always to remove self from NotificationCenter. We don't have object anymore, so it will be handled with 1 line of code for all observables. Removing some specific - when we already handled some logic and we don't need updates from NotificationCenter but object is alive, not deinited and can receive some other notifications, but from another observables

ValentinaIancu-Gini (approved):
@mrkulik Thank you! It looks good!

ValentinaIancu-Gini (approved):
@mrkulik Thanks! Looks good!

PR #718, opened 2024-10-31
BAC-894: send exif data

igor-gini:


ValentinaIancu-Gini (approved):
@igor-gini Looks good to me! Thank you!

PR #720, opened 2024-10-31
Pp 851 reduce build steps

GeraltRiv:


zladzeyka (approved):

.github/workflows/bank-api-library.check.yml
```diff
@@ -71,15 +75,4 @@ jobs:
         CODE_SIGNING_REQUIRED=NO
         ONLY_ACTIVE_ARCH=NO
         CLIENT_ID="gini-mobile-test"
-        CLIENT_SECRET="${{ secrets.GINI_MOBILE_TEST_CLIENT_SECRET }}"
```
zladzeyka, line 74: @GeraltRiv why was it removed it's a part of pinning tests?

zladzeyka: we need to run 1 test separately for  PinningWrongCertificates

zladzeyka (changes requested):
Thank you @GeraltRiv 
1 small change request form my side

.github/workflows/bank-api-library.check.yml
GeraltRiv: Because it is 2 the same `xcodebuild clean test`
I think we just build and test it 2 times

.github/workflows/health-sdk.check.yml
```diff
@@ -64,7 +67,10 @@ jobs:
         ONLY_ACTIVE_ARCH=NO
         CLIENT_ID="gini-mobile-test"
         CLIENT_SECRET="${{ secrets.GINI_MOBILE_TEST_CLIENT_SECRET }}"
-        &&
```
zladzeyka, line 67: please keep the old version we need it for a test with wrong certificate

zladzeyka (changes requested):

zladzeyka (approved):

PR #728, opened 2024-11-13
fix(GiniHealthSDK): IPC-322 distribute gini health sdk as xcframework

igor-gini:


razvancapra (approved):
Nice work 👍

ValentinaIancu-Gini (approved):
@igor-gini I think we can merge this. thank you!

PR #731, opened 2024-11-13
fix(GiniBankSDK): Fix UI bug for "Without skonto" and "With skonto" texts

ValentinaIancu-Gini:
PP-785


mrkulik (approved):
@ValentinaIancu-Gini Great! Thank you for your work!

PR #739, opened 2024-11-21
fix(GiniBankSDK): Fix decimal rounding for Price

mrkulik:
PP-912


ValentinaIancu-Gini (approved):
@mrkulik Thanks for the update. 
Looks good to me!

PR #742, opened 2024-11-25
Release 3.12.0

ValentinaIancu-Gini:


mrkulik (approved):

ValentinaIancu-Gini:
The only check that is not passing is the creation of XCFrameworks.

PR #743, opened 2024-11-25
Ipc 440 import document not uploading multipage and or more than one document

ValentinaIancu-Gini:


razvancapra (approved):
Nice work 👍 
FYI: PageControl view on PaymentReviewViewController isn't working thou. Should we fix that one too?

ValentinaIancu-Gini:
@razvancapra When I tested it was working but it was not visible. The user can swipe through pages. If I missed something then the visibility of the page control please let me know.

PR #744, opened 2024-11-27
IPC-322: Distribute gini health sdk as xcframework

igor-gini:


ValentinaIancu-Gini (approved):
@igor-gini Look good! Thank you for all the effort for making this possible!

PR #753, opened 2024-12-06
PP-777: Add Bank SDK version to upload metadata http header

igor-gini:


BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Document+UploadMetadata.swift
```diff
@@ -81,5 +81,19 @@ extension Document {
             }
             return comment
         }
```
zladzeyka, line 83:
```suggestion
        }
        
```

BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Document+UploadMetadata.swift
```diff
@@ -81,5 +81,19 @@ extension Document {
...
+        static func userComment(_ existingComment: String?, addingIfNotPresent value: String, forKey key: String) -> String {
+            let newValueString = "\(key)=\(value)"
+            guard let existingComment, existingComment.isNotEmpty else {
+                return newValueString
+            }
+            // checking if the key doesn't exist already
+            guard !existingComment
+                .split(separator: ",").map(String.init)
```
zladzeyka, line 91: [Suggestion]: AI brain provides some suggestion here: The issue lies in .split(separator: "="), which operates on each string component after splitting by ",". This chaining of splits and mapping results in a flat array of all substrings but does not correctly preserve the association between keys and values. As a result, the check for key may give false positives or negatives.
Possible solution: Use a dictionary-like approach to handle the split, ensuring keys and values remain paired.

```
// Parse existing comments into key-value pairs
    let keyValues = existingComment
        .split(separator: ",")
        .map { $0.split(separator: "=") }
        .filter { $0.count == 2 }
        .map { (String($0[0]), String($0[1])) }
    
    // Check if the key already exists
    guard !keyValues.contains(where: { $0.0 == key }) else {
        return existingComment
    }
```

BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Document.swift
```diff
@@ -235,6 +235,7 @@ extension Document {
      */
     public struct Metadata {
         var headers: [String: String] = [:]
+        var giniBankVersion: String?
```
zladzeyka, line 238:
```suggestion
        var giniBankSDKVersion: String?
```

BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Document.swift
```diff
@@ -249,29 +250,48 @@ extension Document {
...
+
+            headers[Document.Metadata.headerKeyPrefix + Document.Metadata.uploadHeaderKey] = UploadMetadata.userComment(existingValue, addingIfNotPresent: giniBankVersion, forKey: "GiniBankVer")
+        }
+
         /**
          * Adds upload metadata
          *
-         * - Parameter branchId:            The branch id (i.e: the BLZ of a Bank in Germany)
-         * - Parameter additionalHeaders:   Additional headers for the metadata. i.e: ["customerId":"123456"]
+         * - Parameter uploadMetadata:  Upload datadata
```
zladzeyka, line 287:
```suggestion
         * - Parameter uploadMetadata:  Upload metadata
```

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
```diff
@@ -138,7 +138,14 @@ open class GiniBankNetworkingScreenApiCoordinator: GiniScreenAPICoordinator, Gin
                 api: APIDomain,
                 trackingDelegate: GiniCaptureTrackingDelegate?,
                 lib: GiniBankAPI) {
-        documentService = DocumentService(lib: lib, metadata: documentMetadata)
+        let docMetadata: Document.Metadata = {
```
zladzeyka, line 141: [Suggestion]: Can we extract the new `documentService`code into private function to avoid duplications?

zladzeyka (changes requested):
Thank you @igor-gini please find some suggestions 😇

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
igor-gini: Added a separate function which creates Document.Metadata, DocumentService is not created the same everywhere

BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Document+UploadMetadata.swift
igor-gini: добавил + сделал ещё тесты к usercomment, их не было

igor-gini: Added + created tests for usercomment, there was none

BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Document+UploadMetadata.swift
igor-gini: Github doesn't show for some reason but added a new line there

BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Document.swift
igor-gini: Changed this

BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Document.swift
igor-gini: Fixed this

BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Document.swift
```diff
@@ -249,29 +250,48 @@ extension Document {
...
+        }
+
         /**
          * Adds upload metadata
          *
-         * - Parameter branchId:            The branch id (i.e: the BLZ of a Bank in Germany)
-         * - Parameter additionalHeaders:   Additional headers for the metadata. i.e: ["customerId":"123456"]
+         * - Parameter uploadMetadata:  Upload metadata
          */
         public mutating func addUploadMetadata(_ uploadMetadata: UploadMetadata) {
```
zladzeyka, line 289: [Suggestion]: This is a breaking change can we keep the old method with branchID and additional header, mark it as deprecated and introduce a new one?

zladzeyka (changes requested):

BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Document.swift
igor-gini: I looked through the commit history of this file and it seems that the method with branch ID and additional methods was never present in the first place. The init, which accepts the branch id and additional headers, was always there, and is still there. The documentation for the addUploadMetadata method was wrong(and I fixed it), it likely was overlooked when the upload metadata feature was initially implemented. So short story is: docs were wrong, the method itself hasn't changed, it was like that from the beginning so no new method have been introduced, no old needs to be marked as deprecated..

zladzeyka: thank you @igor-gini

zladzeyka (approved):
Thanks @igor-gini looks good to me!

ValentinaIancu-Gini (approved):
Thanks @igor-gini looks good to me as well!

PR #754, opened 2024-12-09
Pp 615 skonto transfer summary integration tests

mrkulik:


BankSDK/GiniBankSDKExample/Tests/IntegrationTests/TransferSummary/SkontoTransferSummaryIntegrationTest.swift
```diff
@@ -0,0 +1,222 @@
...
+        
+        GiniBankConfiguration.shared.skontoDiscounts = [modifiedSkontoExtractions]
+    }
+
+    override func sendTransferSummary(result: AnalysisResult) {
+        guard let updatedAmountToPayString = result.extractions["amountToPay"]?.value else { return }
+        let amountExtraction = createAmountExtraction(value: updatedAmountToPayString)
+        GiniBankConfiguration.shared.sendTransferSummaryWithSkonto(amountToPayExtraction: amountExtraction,
+                                                                   amountToPayString: updatedAmountToPayString)
+    }
```
ValentinaIancu-Gini, line 206: [clarification] Why do we do this here:
`  guard let updatedAmountToPayString = result.extractions["amountToPay"]?.value else { return }`  and in the same method but for `SkontoNotAppliedTransferSummaryHandler` we don't do this?
If there is no difference can we have just one?

ValentinaIancu-Gini:
@mrkulik Really nice job with these tests!

BankSDK/GiniBankSDKExample/Tests/IntegrationTests/TransferSummary/SkontoTransferSummaryIntegrationTest.swift
mrkulik: @ValentinaIancu-Gini thats true, I've checked that after some changes (during development) I have same overrides for both specific handlers and they are the same as default implementation in base handler. I removed this and left only base implementation for both, thank you for raising it!

ValentinaIancu-Gini (approved):
@mrkulik Thank you!

PR #755, opened 2024-12-09
fix(GiniBankSDK): Expose cancel sdk function 

igor-gini:
PP-129

### DIFF
--- a/BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Core/GiniBankAPI.swift
+++ b/BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Core/GiniBankAPI.swift
@@ -7,7 +7,10 @@
 
 import Foundation
 
-// should only be used in tests
+/**
+Sets the default error logger. It is only used when giniErrorLoggerIsOn is true.
+- note: Internal usage only.
+**/
 var _GINIBANKAPILIBRARY_DISABLE_KEYCHAIN_PRECONDITION_FAILURE: Bool = false
 
 /// The Gini Bank API Library
--- a/BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Core/GiniBankAPI.swift
+++ b/BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Core/GiniBankAPI.swift
@@ -7,12 +7,6 @@
 
 import Foundation
 
-/**
-Sets the default error logger. It is only used when giniErrorLoggerIsOn is true.
-- note: Internal usage only.
-**/
-var _GINIBANKAPILIBRARY_DISABLE_KEYCHAIN_PRECONDITION_FAILURE: Bool = false
-
 /// The Gini Bank API Library
 public final class GiniBankAPI {
     
@@ -151,10 +145,8 @@ extension GiniBankAPI {
                                                                    value: client.domain,
                                                                    service: .auth))
             } catch {
-                if !_GINIBANKAPILIBRARY_DISABLE_KEYCHAIN_PRECONDITION_FAILURE {
-                    preconditionFailure("There was an error using the Keychain. " +
-                                        "Check that the Keychain capability is enabled in your project")
-                }
+                preconditionFailure("There was an error using the Keychain. " +
+                                    "Check that the Keychain capability is enabled in your project")
             }
         }
     }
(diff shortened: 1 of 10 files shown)


BankSDK/GiniBankSDKExample/GiniBankSDKExample/DemoViewController.xib
```diff
@@ -117,26 +118,37 @@
                     <nil key="textColor"/>
                     <nil key="highlightedColor"/>
                 </label>
+                <stackView opaque="NO" contentMode="scaleToFill" spacing="16" translatesAutoresizingMaskIntoConstraints="NO" id="QRX-Mq-lHn">
```
ValentinaIancu-Gini, line 121: [clarification]Why did we not add this new option in the Settings screen? We will have more things to fix when we support landscape mode and as well in the Photo payment team we will add a new button to see the "transaction list". I would suggest moving it inside the Settings screen at the button of the screen where all the "DEbug and development options" are located.

ValentinaIancu-Gini (commented):
@igor-gini I finished the code review for this pr. Can you please check my suggestions and questions? Thank you!

zladzeyka (changes requested):
@igor-gini please add also a unit test

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
```diff
@@ -15,6 +15,7 @@ protocol Coordinator: AnyObject {
 }
 
 open class GiniBankNetworkingScreenApiCoordinator: GiniScreenAPICoordinator, GiniCaptureDelegate {
+    static var currentCoordinator: GiniBankNetworkingScreenApiCoordinator?
```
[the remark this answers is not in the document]
igor-gini: This needs to be at least at internal level so tests would be able to access it.

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/Networking/GiniBank+Networking.swift
```diff
@@ -10,6 +10,12 @@ import UIKit
 import GiniCaptureSDK
 import GiniBankAPILibrary
 extension GiniBank {
+    /**
+     Aborts the current GiniBankSDK session
```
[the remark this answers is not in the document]
igor-gini: Sure, added

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
```diff
@@ -284,6 +286,15 @@ open class GiniBankNetworkingScreenApiCoordinator: GiniScreenAPICoordinator, Gin
         })
         return self.start(withDocuments: documents, animated: animated)
     }
+
+    public func closeSDK() {
```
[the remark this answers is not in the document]
igor-gini: I made it public, so a particular instance could be cancelled from anywhere, but I suppose it’s not actually needed, replaced it with calling finishWithCancellation directly, followed by cleaning up GiniBankConfiguration(ScreenAPICoordinator seems to be doing this in closeSreenAPIAndSendTransferSummary, so I guess we also need that here).

BankSDK/GiniBankSDKExample/GiniBankSDKExample/DemoViewController.xib
igor-gini: Added a Settings item

igor-gini:
> @igor-gini please add also a unit test

I added a test, but it passes only if GiniBankAPI does not trigger preconditionFailure that there is no entitlement for access to Keychain. I added a possible solution: I made the global internal variable _GINIBANKAPILIBRARY_DISABLE_KEYCHAIN_PRECONDITION_FAILURE and set it to true in the test.

BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Core/GiniBankAPI.swift
```diff
@@ -7,6 +7,9 @@
 
 import Foundation
 
+// should only be used in tests
```
zladzeyka, line 10: [Suggestion] Please add a comment with more explanation like:

```
 /**
     Sets the default error logger. It is only used when giniErrorLoggerIsOn is true.

     - note: Internal usage only.
   */
```

zladzeyka (changes requested):
Thank you for the update @igor-gini 
2 tiny suggestions from my side :)

BankAPILibrary/GiniBankAPILibrary/Sources/GiniBankAPILibrary/Documents/Core/GiniBankAPI.swift
igor-gini: Sure, done

zladzeyka (approved):

PR #761, opened 2025-01-07
Gini bank sdk xc frameworks bash script

ValentinaIancu-Gini:
This pr dosen't take into consideration Pinning SDKs with the TrustKit framework.


ValentinaIancu-Gini:
I will close this pr and open a new one when needed.

PR #768, opened 2025-01-17
fix(GiniHealthSDK): Fix numpad keyboard for changing amount

razvancapra:
IPC-546


zladzeyka (approved):
Thank you @razvancapra

PR #769, opened 2025-01-20
ci: Fix search for the ticket number

zladzeyka:
I fixed the search for a ticket number in the script and https://ginis.atlassian.net/browse/PP-74 script was working
I manually triggered the workflow in the branch and build was added in Jira.
PP-74


ValentinaIancu-Gini (approved):
@zladzeyka very nice!

PR #770, opened 2025-01-21
Ipc 551 fix delegate

zladzeyka:
- Add tracking event  closebuttonClicked
- Fixed delegate event  didCreatePaymentRequest(paymentRequestId: String)
- Fixed delegate set in init(giniApiLib: GiniHealthAPI) used in init with AlternativeTokenSource


razvancapra (approved):
Nice job ✅

PR #785, opened 2025-02-21
Pp 926 implement user journey analytics for skonto

ValentinaIancu-Gini:
Add document_type as a super property and remove it where was set for specific events.


mrkulik (approved):
@ValentinaIancu-Gini 👍

PR #789, opened 2025-03-06
GitHub action workflows improvements

ValentinaIancu-Gini:
- the `edited` type was removed because we don't want to trigger the flows when the pr title or description is modified.
- `tags-ignore` does not apply to `workflow_dispatch` or `pull_request`


zladzeyka (approved):
🥳

PR #791, opened 2025-03-07
fix(GiniCaptureSDK): Use unexpected error for 404 upload failure

ValentinaIancu-Gini:
When a 404 error occurs on the invoice preview screen, the error message displayed is misleading. It does not correctly convey that the document is unavailable, which may cause confusion for users.
- decision: Use unexpected error for 404 upload failure
- 
PP-998


mrkulik (approved):
@ValentinaIancu-Gini 🙌

zladzeyka (approved):

PR #793, opened 2025-03-10
Pp 1047 t ds attach a file first and then an image is displaying previously selected file s name for the image

ValentinaIancu-Gini:


mrkulik (approved):
@ValentinaIancu-Gini 👍

PR #797, opened 2025-03-12
Pm 151 jira ticket extraction improvements

obaidgini:
Improvement for extraction of Jira ticket 
PM-151


zladzeyka (approved):
@obaidgini thank you, let's add in our backlog a ticket for checking how can we improve those workflows and avoid duplications.

PR #800, opened 2025-03-14
ci: Fix jira ticket extraction when we don't find any ticket in the last 3 commits

ValentinaIancu-Gini:
This fix will allow the job to pass even if there is no ticket found in the last 3 commits.


razvancapra (approved):
🤞

obaidgini (approved):
Looks good 👌

PR #815, opened 2025-03-24
Try fixing deprecated ad-hoc export option

igor-gini:


ValentinaIancu-Gini (approved):
@igor-gini If this is the only change, for me is looking good.
Did you tried to trigger upload to Firebase and check if the warning is solved?

igor-gini:
> @igor-gini If this is the only change, for me is looking good. Did you tried to trigger upload to Firebase and check if the warning is solved?

Yes :)

https://github.com/gini/gini-mobile-ios/actions/runs/14054818491/job/39351876102

Look into between `ARCHIVE SUCCEEDED` and `EXPORT SUCCEEDED` and there is no more deprecated message

PR #816, opened 2025-03-24
Pp 1049 i os error title and description doesn t match requirements

ValentinaIancu-Gini:


mrkulik (approved):
@ValentinaIancu-Gini thank you for helping with this!

PR #819, opened 2025-03-28
Pp 1065 i os remove deprecated public methods

mrkulik:


ValentinaIancu-Gini (approved):
@mrkulik thank you for your great work!

PR #824, opened 2025-04-03
Pp 1081 move gini color scheme inside the gini utilities sdk

mrkulik:


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Resources/GiniCaptureColorScheme.swift
```diff
@@ -0,0 +1,185 @@
+//
```
ValentinaIancu-Gini, line 1: [Suggestion] Let's remove this since we did not start using it, yet. I would also like to plan something that will fit better with the new system design and the other products that we have.

ValentinaIancu-Gini (approved):
@mrkulik The code looks good. I will ask Vlad to test it before merging it.

ValentinaIancu-Gini:
I'm merging this and adding a new ticket for the testing this in a separate ticket.

PR #831, opened 2025-04-11
Pp 1114 i os implement the new order of the ct as in error screen

mrkulik:


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Error/ErrorScreenViewController.swift
```diff
@@ -21,10 +21,10 @@ class ErrorScreenViewController: UIViewController {
 
     lazy var buttonsView: ButtonsView = {
```
ValentinaIancu-Gini, line 22: [Suggestion] We should make this private.

ValentinaIancu-Gini (commented):
@mrkulik I left some suggestions. 
As well same suggestion to make private `configuration` property inside the screens and would be nice to check what else can we make private. Thank you!

ValentinaIancu-Gini (approved):
@mrkulik Thanks for the updates! The PR is looking good! Let's ask Vlad to test it.

ValentinaIancu-Gini:
I will merge this PR since the failing test is not caused by the changes in this PR. To fix the test, I will open a new pr directly in the main branch.

PR #837, opened 2025-04-17
fix(GiniBankSDKExample): Fix integration test for payment request

ValentinaIancu-Gini:
Fix for the payment request test that is failing in the bank SDK tests.


lestergini (approved):
Thanks for the code cleaning @ValentinaIancu-Gini. Code looks good to me.

mrkulik (approved):
@ValentinaIancu-Gini thank you so much for fixing it!

PR #840, opened 2025-04-22
fix(GiniBankSDKExample): Fix integration test for instant payment feature when the default value is false

ValentinaIancu-Gini:
Add a missing condition for the default value of instant payment.
For now, for every case that is not instant payment detected on the invoice or is on the invoice but not check marked, we receive from CVIE it as false


mrkulik (approved):
@ValentinaIancu-Gini tests are working now, thank you for resolving it!

PR #854, opened 2025-05-05
Pp 1182 page controller issue

qnaveed87:


zladzeyka:
We discussed together that we keep the native behaviour.

PR #859, opened 2025-05-09
Pp 1221 implement local storage for the client configuration flag and logic to display the message

ValentinaIancu-Gini:
Implement local storage for the client configuration flag and logic to display the message for the capturing invoice flow.


mrkulik (approved):
@ValentinaIancu-Gini thank you!

PR #861, opened 2025-05-12
Fix qrcode flow navigation

ValentinaIancu-Gini:
In this pr the navigation is fixed when a QR code is scanned successfully.


mrkulik (approved):
@ValentinaIancu-Gini thank you for fixing it!

PR #871, opened 2025-05-19
Pp 1261 bottom bar height issue

qnaveed87:


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Error/ErrorScreenViewController.swift
```diff
@@ -91,6 +92,11 @@ class ErrorScreenViewController: UIViewController {
 
         sendAnalyticsScreenShown()
     }
+    
+    override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+        navigationBarHeightConstraint.constant = getBottomBarHeight()
+    }
```
ValentinaIancu-Gini, line 99: [clarification] @qnaveed87 is it really necessary to use this method? In the Skonto help screens I used 
```
    override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
        super.viewWillTransition(to: size, with: coordinator)
```
can we use this method instead of `viewDidLayoutSubviews` which is super "expensive" ?

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Error/ErrorScreenViewController.swift
```diff
@@ -54,6 +54,7 @@ class ErrorScreenViewController: UIViewController {
     private var navigationBarBottomAdapter: ErrorNavigationBarBottomAdapter?
     private var buttonsHeightConstraint: NSLayoutConstraint?
     private var buttonsBottomConstraint: NSLayoutConstraint?
+    private var navigationBarHeightConstraint: NSLayoutConstraint = NSLayoutConstraint()
```
ValentinaIancu-Gini, line 57: [clarification] Can we make this a lazy var?

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Error/ErrorScreenViewController.swift
```diff
@@ -334,6 +345,18 @@ class ErrorScreenViewController: UIViewController {
         static let sidePadding: CGFloat = 24
         static let iPadWidthMultiplier: CGFloat = 0.7
         static let iPadButtonsWidth: CGFloat = 280
-        static let navigationBarHeight: CGFloat = 114
+        static let navigationBarHeight: CGFloat = 110
+        static let navigationBarHeightLandscape: CGFloat = 64
```
ValentinaIancu-Gini, line 349: [Clarification] I checked Figma and the height for the bar should be 62 on big iPhones and on the small ones should be 40.

qnaveed87: Needs discussion

qnaveed87: @ValentinaIancu-Gini for this one we need to create a ticket and mention which screen should be updated.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Error/ErrorScreenViewController.swift
```diff
@@ -48,12 +48,22 @@ class ErrorScreenViewController: UIViewController {
         scrollView.translatesAutoresizingMaskIntoConstraints = false
         return scrollView
     }()
+    
+    lazy var navigationBarHeightConstraint: NSLayoutConstraint? = {
```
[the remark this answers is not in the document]
qnaveed87: Yes we can. I missed it 
Thanks :)

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Error/ErrorScreenViewController.swift
```diff
@@ -187,13 +203,14 @@ class ErrorScreenViewController: UIViewController {
...
 
+
         NSLayoutConstraint.activate([
             buttonsView.bottomAnchor.constraint(equalTo: navigationBar.topAnchor,
                                                 constant: -GiniMargins.margin),
             navigationBar.bottomAnchor.constraint(equalTo: view.bottomAnchor),
             navigationBar.leadingAnchor.constraint(equalTo: view.leadingAnchor),
             navigationBar.trailingAnchor.constraint(equalTo: view.trailingAnchor),
-            navigationBar.heightAnchor.constraint(equalToConstant: Constants.navigationBarHeight)
+            navigationBarHeightConstraint!
```
ValentinaIancu-Gini, line 213: [Clarification] If tthis really needed to be force unwrapped? We should avoid force-unwrapping. I suggest to add an if statement and add this constraint only when is not nil.

qnaveed87: This constraint will only active if we have bottom bar and hence it cant be nil. But for the safe side i will add if statement 

`if let navigationBar = navigationBarBottomAdapter?.injectedView() {
                bottomNavigationBar = navigationBar
                navigationBar.translatesAutoresizingMaskIntoConstraints = false
                view.addSubview(navigationBar)

                layoutBottomNavigationBar(navigationBar)
            }`
Thanks

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Error/ErrorScreenViewController.swift
```diff
@@ -48,12 +48,22 @@ class ErrorScreenViewController: UIViewController {
...
+        let constraint = navbar.heightAnchor.constraint(equalToConstant: getBottomBarHeight())
+        return constraint
+    }()
 
     let viewModel: BottomButtonsViewModel
     private let errorType: ErrorType
     private var navigationBarBottomAdapter: ErrorNavigationBarBottomAdapter?
     private var buttonsHeightConstraint: NSLayoutConstraint?
     private var buttonsBottomConstraint: NSLayoutConstraint?
+    private var bottomNavigationBar: UIView?
```
ValentinaIancu-Gini, line 65: [clarification] I don't understand why we need this var.

qnaveed87: `lazy var navigationBarHeightConstraint: NSLayoutConstraint? = {
        guard let navbar = bottomNavigationBar else {
            return nil
        }
        let constraint = navbar.heightAnchor.constraint(equalToConstant: getBottomBarHeight())
        return constraint
    }()`

If the view has bottom bar than above constraint set height anchor to it.

PR #883, opened 2025-05-26
PP-1347 Revert change for external keyboard

ValentinaIancu-Gini:
PP-1347


lestergini (approved):
Great catch @ValentinaIancu-Gini, I don't have comments to add.

PR #888, opened 2025-06-02
Improvements githubflow gini bank sdk

ValentinaIancu-Gini:
This pull request introduces a sequential chain of GitHub Actions workflows for processes related to the GiniBankSDK.

Previously, all workflows were triggered in parallel, which could lead to unnecessary execution of dependent flows even if a preceding one failed. With this update:

- Workflows are now executed sequentially rather than in parallel.
- Subsequent workflows are cancelled if the preceding one fails, ensuring that only successful builds trigger downstream flows.

This change improves efficiency, reduces resource waste, and increases clarity in the CI process.

### DIFF
--- a/.github/workflows/bank-api-library.build.docs.yml
+++ b/.github/workflows/bank-api-library.build.docs.yml
@@ -1,6 +1,7 @@
 name: Builds docs for Bank API Library
 
 on:
+  workflow_call:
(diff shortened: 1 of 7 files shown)


ValentinaIancu-Gini:
I will close for now this pr. I will open a new one once there is a need for this.

PR #898, opened 2025-06-11
PP-1239-Disable-bottom-tips-when-the-QR-code-education-message-is-displayed-on-the-Analysis-screen

ValentinaIancu-Gini:
When the QR code education message is displayed, the bottom screen tips should not be displayed. In case the request is slow, in landscape mode these tips are hidding the qrcode education message.


mrkulik (approved):
@ValentinaIancu-Gini 👍 looks great!

PR #907, opened 2025-06-16
fix(GiniBankSDKExample): Update the stage api endpoint

ValentinaIancu-Gini:
In this PR, the API endpoint for the stage environment is updated.


qnaveed87 (approved):
Looks good to me

zladzeyka (approved):

PR #920, opened 2025-06-23
fix(GiniCaptureSDK): layout issue on if device starts with landscape

qnaveed87:
PP-1358


qnaveed87:
Thank you  @zladzeyka . UI have been handled by viewWillTransition(to:with:) this is only for the use case when users enters the screen in landscape mode. In this case viewWillTransition(to:with:) wont called. Before it were handled by viewWillAppear but that was too early for the view and hence UI does not align properly.

zladzeyka (approved):
thanks @qnaveed87

ValentinaIancu-Gini (approved):
@qnaveed87 Thank you! Look good to me!

PR #930, opened 2025-07-04
Ipc 695 keyboard issue

qnaveed87:
## Pull Request Description

This PR fixes the issue on the Payment Review screen without a document where keeping the keypad open and changing to portrait mode causes the UI to shift unexpectedly.

[IPC-695](https://ginis.atlassian.net/browse/IPC-695)

I fixed the issue by calling `reset` in the `keyboardWillHide` event to force the layout to update correctly after orientation changes.

## Notes for Reviewers

- I changed the unsubscribe event to be called from `deinit` instead of `viewDidDisappear` to ensure event listeners are properly removed only when the view controller is fully deallocated.
- Cleaned up trailing whitespace from the code for better formatting.

**Devices tested:** iPhone X (Real Device), iPhone 11 (Real Device), iPhone SE 1st Gen (Simulator)


ValentinaIancu-Gini (approved):
@qnaveed87, from a technical perspective looks good to me. Thank you!

PR #972, opened 2025-08-26
Ipc 788 duplicated payment providers

lestergini:
## Pull Request Description

This PR adds the following:

- A call to `didDismissHealthSDK` for the cases when an error is presented before navigating to `PaymentReviewViewController`
- Add the `uniqued` method to an `Array` extension to prevent duplications on `paymentProviders` list.
- Removes the `is_stable_release` validation from `publish_docs` lane to be able to make a release of the documentation from any branch. Not only `main`

[IPC-788]

## Notes for Reviewers

- In order to test the error scenario, you need to use Charles to edit the request of the `get /documents` endpoint to use an invalid id and get a 404. Also need to enable the `alternative navigation` flag.

### DIFF
--- a/GiniComponents/GiniUtilites/Sources/GiniUtilites/Extensions/Array+Extensions.swift
+++ b/GiniComponents/GiniUtilites/Sources/GiniUtilites/Extensions/Array+Extensions.swift
@@ -1,0 +1,31 @@
+//
+//  Array+Extensions.swift
+//
+//  Copyright © 2025 Gini GmbH. All rights reserved.
+//
+
+
+extension Array {
+    
+    /**
+     * Removes duplicate elements from the array based on a specified key.
+     *
+     * This method maintains the order of elements, keeping the first occurrence
+     * of each unique element and removing subsequent duplicates.
+     *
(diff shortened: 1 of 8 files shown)


GiniComponents/GiniUtilites/Sources/GiniUtilites/Extensions/Array+Extensions.swift
```diff
@@ -0,0 +1,31 @@
...
+     *   specified key, preserving the original order.
+     *
+     * Example:
+     * ```swift
+     * let paymentProviders = [provider1, provider2, provider1, provider3]
+     * let uniqueProviders = paymentProviders.uniqued(by: { $0.id })
+     * ```
+     */
+    public func uniqued<T: Hashable>(by transform: (Element) -> T) -> [Element] {
+        var seen = Set<T>()
```
qnaveed87, line 28: [Suggestion]: How about renamed it something like seenArray or seenKeys ?

qnaveed87 (commented):
Looks good just a minor naming suggestion

PR #985, opened 2025-09-29
Pp 1566 sonar qube add a nested comment explaining why the function is empty or complete the implementation

ValentinaIancu-Gini:
## Pull Request Description

[PP-1566](https://ginis.atlassian.net/browse/PP-1566)


BankSDK/GiniBankSDKExample/GiniBankSDKExample/Views/OnButtonLoading.swift
```diff
@@ -0,0 +1,30 @@
...
+        self.startAnimating()
+    }
+
+    public func stopAnimation() {
+        self.stopAnimating()
+    }
+
+    public func injectedView() -> UIView {
+        self.style = .large
+        self.color = .red
```
qnaveed87, line 21: [Suggestion]: How about using it via configurations instead of hardcoded color ?

qnaveed87 (commented):
@ValentinaIancu-Gini looks good to me. I just have small suggestion. Kindly can you look at it?

BankSDK/GiniBankSDKExample/GiniBankSDKExample/Views/OnButtonLoading.swift
ValentinaIancu-Gini: @qnaveed87, this is just an example in the demo app. I would not complicate things and at least not in this pr. I just moved the code from one file to a separate file for sonarqube to not complain of having to separate objects in the same file.
Is that ok for you?

qnaveed87 (approved):
@ValentinaIancu-Gini thank you for the clarification

PR #1100, opened 2026-04-15
Pp 2483 enable UI test files app simulators

qnaveed87:
# Pull Request Description

This PR makes the **BankSDK example app's Files-picker UI tests runnable on the iOS Simulator**.

Previously, these tests only worked on real devices (via BrowserStack) because the simulator's **Files** app was always empty—there was no invoice PDF under:

> **On My iPhone → GiniBankSDKExample**

for the test to select.

## How it works

Every UI test now starts by running a small Swift helper named `copyFixturesToSimulator` (located in the UITests target's base class).

Before the app is launched, this helper:

1. Locates the app container that Xcode created for the test run inside the simulator.
2. Copies the PDF invoice fixtures from:

   ```
   GiniBankSDKExampleUITests/TestSamples/TestSamplesForBS/
   ```

3. Places them into the app's Documents directory, making them visible in the Files app under:

   ```
   Browse → On My iPhone → GiniBankSDKExample
   ```

## UI Test Flow

Once the app is running, the UI test performs the following navigation:

```
Browse
  → On My iPhone
      → GiniBankSDKExample
          → Select invoice PDF
              → Open
```

This flow is implemented in `MainScreen.tapFileWithName`.

### What changed?

The previous implementation searched in **Recents**, which is no longer reliable on modern iOS versions.

The updated implementation:

- Navigates directly to **Browse → On My iPhone → GiniBankSDKExample**
- Taps the **file cell** instead of the filename label

Selecting the cell ensures the Files picker treats the action as a file selection rather than opening a preview.

## How to try locally

> **Important:** Make sure **only one iOS Simulator is booted** before running the tests.

If multiple simulators are running, the helper may copy the fixtures into the wrong simulator's app container.

This happens because:

- Xcode launches the app on the selected destination simulator.
- The helper searches the simulator where the test runner is installed.
- With multiple booted simulators, these may not be the same device.

You can shut down extra simulators using either:

- **Xcode → Window → Devices and Simulators**, or
- Terminal:

```bash
xcrun simctl shutdown all
```


ValentinaIancu-Gini (approved):
Looks good to me!

PR #1113, opened 2026-04-27
Fix/heal 336 unresponsive payment review

zladzeyka:
## Pull Request Description

After dismissing the bank selection sheet, the payment review screen intermittently became unresponsive to all touch input. The overlay   
  UIWindow (at windowLevel = .normal + 1) remained visible for one run loop cycle after dismissal — necessary to avoid a UIKit crash
  mid-transition — but its root view was a plain UIView that silently consumed all touches during that gap.                                 
                                                            
  How

  Introduced GiniPassthroughView, a UIView subclass that overrides hitTest(_:with:) to return nil when the hit view is its own background,  
  letting touches fall through to the main window. Touches on the presented sheet resolve to subviews and are unaffected. Set as the root
  view of GiniPassthroughViewController via loadView().                                                                                     
                                                            
  Notes for Reviewers

  How to verify:                                                                                                                            
  1. Open the Payment Review screen.
  2. Tap the bank selector button.                                                                                                          
  3. Select a different bank from the list and dismiss the sheet.
  4. Interact with any button on the payment review screen.      
                                                                                                                                            
  Before: Screen fully unresponsive after dismissal, no console output.                                                                     
  After: All buttons respond normally.                                                                                                      
                                                                                                                                            
  The fix is one small class — no logic changes to presentation or teardown. The deferred window.isHidden = true in tearDown() is           
  intentionally kept to prevent EXC_BAD_ACCESS mid-transition; GiniPassthroughView handles the touch side effect of that deferral.
  
  [HEAL-336](https://ginis.atlassian.net/browse/HEAL-336)

[HEAL-336]: https://ginis.atlassian.net/browse/HEAL-336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/GiniComponents/InternalPaymentSDK/GiniInternalPaymentSDK/Sources/GiniInternalPaymentSDK/PaymentReview/GiniOverlayWindowPresenter.swift
+++ b/GiniComponents/InternalPaymentSDK/GiniInternalPaymentSDK/Sources/GiniInternalPaymentSDK/PaymentReview/GiniOverlayWindowPresenter.swift
@@ -93,6 +93,20 @@ final class GiniOverlayWindowPresenter {
 }
 
 /**
+ A UIView subclass that passes touches through when nothing meaningful is hit.
+
+ When no presented sheet is covering the screen, `hitTest` returns `nil` so that
+ touch events fall through to the underlying main window instead of being silently
+ consumed by the transparent overlay window.
+ */
+private final class GiniPassthroughView: UIView {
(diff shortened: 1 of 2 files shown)


ValentinaIancu-Gini (approved):
@zladzeyka looks good. Thank you!

PR #1126, opened 2026-05-05
fix(GiniInternalPaymentSDK): Restore close and back button interaction on iOS 16

zladzeyka:
## Problem

On iOS 16.0–16.3, the close (X) button and back button on the Payment Review screen are unresponsive after selecting a bank via the Transfer Directly flow.

**Steps to reproduce (iOS 16 only):**
1. Open example app → tap "Start with test document"
2. Tap "Transfer directly"
3. Select a bank → arrive at Payment Review screen
4. Tap the close (X) button or back button → nothing happens

**Root cause:** `presentationBackgroundInteraction` (the SwiftUI API that allows touches to pass through to the presenting view behind a sheet) was introduced in iOS 16.4. `GiniBottomSheetModifier` already applies it on iOS 16.4+, but the `else` branch for iOS < 16.4 returned the sheet unmodified. iOS's default sheet behaviour on 16.0–16.3 dims and completely blocks touches on the presenting view, making navigation-bar buttons unreachable.

## Fix

In `GiniBottomSheetModifier`, for iOS < 16.4, embed a `UIViewControllerRepresentable` helper as the sheet content's `.background`. Its `UIViewController` sets `sheetPresentationController?.largestUndimmedDetentIdentifier = .large` in `viewWillAppear` — the UIKit equivalent that leaves the background undimmed and fully interactive. iOS 16.4+ continues to use `.presentationBackgroundInteraction` unchanged.

## Changes

- `GiniBottomSheetModifier.swift` — add `SheetBackgroundInteractionHelper` (UIViewControllerRepresentable) applied in the iOS < 16.4 else branch

## Test plan

- [ ] Run on iOS 16 simulator: open example app → "Start with test document" → "Transfer directly" → select bank → verify close (X) and back buttons are tappable
- [ ] Run on iOS 16.4+ / iOS 17 / iOS 18: verify existing behaviour unchanged
- [ ] Run `GiniInternalPaymentSDK` unit tests — all pass

HEAL-355

### DIFF
--- a/GiniComponents/InternalPaymentSDK/GiniInternalPaymentSDK/Sources/GiniInternalPaymentSDK/BottomSheet/GiniBottomSheetModifier.swift
+++ b/GiniComponents/InternalPaymentSDK/GiniInternalPaymentSDK/Sources/GiniInternalPaymentSDK/BottomSheet/GiniBottomSheetModifier.swift
@@ -78,7 +78,7 @@ struct SheetBackgroundInteractionHelper: UIViewControllerRepresentable {
         SheetBackgroundInteractionHelperController()
     }
 
-    func updateUIViewController(_ controller: SheetBackgroundInteractionHelperController,
+    func updateUIViewController(_: SheetBackgroundInteractionHelperController,
                                 context _: Context) {
         // No updates needed — configuration is applied once in viewWillAppear.
     }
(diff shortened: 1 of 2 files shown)


zladzeyka:
We decided to raise SDK target and skip this edge case bug for iOS 16.0-16.3

PR #1136, opened 2026-05-11
fix(GiniInternalPaymentSDK): Fix Dynamic Type not responding to runtime font size changes

zladzeyka:
[HEAL-344](https://ginis.atlassian.net/browse/HEAL-344)

Fixes Dynamic Type support across the payment SDK so all UI scales correctly at every accessibility font size and responds to runtime size changes.

### What changed

**FontProvider** (`GiniUtilites`)
- Simplified to compute fonts fresh on every `font(for:)` call via `UIFontMetrics` at call time, so fonts always reflect the current Dynamic Type size without storing pre-scaled instances.

**TextFieldConfiguration** (`GiniUtilites`)
- Removed the `dynamicTextFont` closure and `resolvedTextFont` shim — no longer needed since `FontProvider.font(for:)` always returns the current scaled font.

**BanksBottomView** (`GiniInternalPaymentSDK`)
- At accessibility font sizes (AX1+), title and description move into a `tableHeaderView` so they scroll with the bank list instead of being clipped above it.
- Fixed a crash on bank picker tap: `FontProvider` already returns a `UIFontMetrics`-scaled font; wrapping it again in `UIFontMetrics.default.scaledFont(for:)` triggered *"A scaled font returned by UIFontMetrics should not be used to create another scaled font."* Removed all double-scaling wrappers.
- Fixed `tableHeaderView` height measurement: uses `preferredMaxLayoutWidth` on labels before calling `systemLayoutSizeFitting` so description text wraps correctly at any screen width.
- Fixed `estimatedRowHeight` to `UITableView.automaticDimension` (was a fixed 64 pt constant that badly underestimated cell height at large font sizes).
- Removed memory leak: `UIContentSizeCategory.didChangeNotification` observer was added in `viewDidLoad` but never removed; now removed in `deinit`.

**BankSelectionTableViewCell** (`GiniInternalPaymentSDK`)
- Removed double `UIFontMetrics` scaling on `bankNameLabel`.

**ShareInvoiceBottomView** (`GiniInternalPaymentSDK`)
- Removed double `UIFontMetrics` scaling on title, description, and create labels.
- Applied the same accessibility layout pattern as `BanksBottomView`.

**MoreInformationView** (`GiniInternalPaymentSDK`)
- Limited "More information." label to one line with `minimumScaleFactor` so the bottom bar does not grow excessively at large font sizes.

**PaymentInfoViewController / PaymentInfoViewModel** (`GiniInternalPaymentSDK`)
- Added `UIContentSizeCategory.didChangeNotification` observer to rebuild attributed strings and reload the FAQ table on runtime font size changes.
- Removed the 100 pt height cap on `payBillsTitleLabel` that clipped the title at AX sizes.

### Notes for Reviewers

- Verified on iPhone 17 simulator at AX5 (largest accessibility size) and default size.
- Verified bank picker tap no longer crashes at any font size.
- Description label in bank selection sheet now wraps to multiple lines at large sizes.
- Normal (non-accessibility) layouts are untouched — all accessibility changes are guarded by `UIAccessibility.isAccessibilityCategory`.
- Unit tests updated: removed dead assertion on value-type `ClientConfiguration`, removed test for deleted `dynamicTextFont` property, fixed `giniDynamicFont` test to use SDK text styles (`.body1`, `.body2`, `.button`).
- Rotation from landscape to portrait re-sizes the header correctly via `viewWillTransition`.
- No known limitations. Follow-up: verify first-presentation header height on older iOS versions if needed.

[HEAL-344]: https://ginis.atlassian.net/browse/HEAL-344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ValentinaIancu-Gini:
@zladzeyka I tested but we still have a few issues. I will report them here and see if we need to create a new ticket for them
IMG_0063
IMG_0712
IMG_0713
IMG_0714
IMG_0716

https://github.com/user-attachments/assets/ea4262f2-f48b-497b-81a3-c8abed3fbd11

ValentinaIancu-Gini:
@zladzeyka I tested this on iPhone X with iOS 16.7.15.

GiniComponents/Utilities/GiniUtilites/Sources/GiniUtilites/Font/FontProvider.swift
```diff
@@ -9,10 +9,10 @@ import UIKit
 import SwiftUI
 
 public final class FontProvider {
-    private var textStyleFonts = FontProvider.defaultFonts
+    private var textStyleFonts: [UIFont.TextStyle: UIFont] = [:]
 
     public init() {
-        //empty initializer for public acces
+        // Empty initializer for public access.
```
qnaveed87, line 15: [Suggestion] Can we please do it like this 
`/**
  Public initializer for SDK consumers.
  */`
` public init() {}`

qnaveed87 (commented):
Thank you for the good work @zladzeyka 

I have left one small comment

GiniComponents/Utilities/GiniUtilites/Sources/GiniUtilites/Font/FontProvider.swift
ValentinaIancu-Gini: @qnaveed87 no, we need a comment inside the empty method to satisfy Sonar Cloud requirment.

ValentinaIancu-Gini:
@zladzeyka I restested my scenarios and if I first enable accessibiity and then go with testing the app from the beginning, I don't see any more of these issues. I would propose to merge this PR.

ValentinaIancu-Gini (approved):
@zladzeyka the merge conflict was fixed. Let's merge and see what is remaning to be fixed.

PR #1140, opened 2026-05-12
fix(GiniInternalPaymentSDK): Animate and reposition info banner on Payment Review sheet

zladzeyka:
## Pull Request Description

Fixes the green info banner on the Payment Review screen.

The banner was rendered in a `ZStack` overlay at the top of the sheet, which caused two problems:
- It covered the Recipient field, hiding pre-filled data
- The sheet height (`contentHeight`) only measured the form fields, so when the banner appeared the buttons at the bottom were pushed out of view

### How it was implemented

The banner is now the first item inside the fields `VStack` within `baseScrollView`. Because `getHeight` is applied to that `VStack`, it naturally includes the banner height and the sheet's `presentationDetents` detent grows to fit all content when the banner slides in.

The entry animation was changed from a manual `offset(y:) + opacity` driven by a separate `bannerVisible` state to SwiftUI's `.transition(.move(edge: .top).combined(with: .opacity))`. The `ScrollView` clips the view as it slides in from the top, producing the same slide-down effect without needing `.clipped()` on a wrapping `ZStack`.

The corner radius was updated from top-only to all corners, since the banner is no longer pinned to the top edge of the sheet. `bannerVerticalPadding` was reduced from 16 to 12 pt for a more compact appearance.

**Affected module:** `GiniInternalPaymentSDK` — `PaymentReviewPaymentInformationView`, `PaymentReviewObservableModel` (whitespace only)

## Notes for Reviewers

**Test scenario:**

1. Open the Payment Review screen with a pre-filled invoice
2. Observe the green "Please check the pre-filled data." banner slides down from the top of the sheet (not from behind the grabber) with a spring animation
3. All four form fields (Recipient, IBAN, Amount, Reference number) and the bank picker + "To the banking app" button are fully visible below the banner
4. After the configured popup duration (~3 s) the banner slides back up; the form snaps back to its full height
5. Verify on both small (iPhone Air) and large (iPhone 17 Pro Max) simulators — buttons must not be cut off on either

**Devices tested:** iPhone Air (iOS 18.5), iPhone 17 Pro Max (iOS 18.5)

**Known limitation:** The sheet resize animation (sheet growing when banner appears) is driven by the `presentationDetents` change and may not be perfectly in sync with the banner's spring animation on iOS 16. Functionally correct on iOS 17+.

No new unit tests were added — the change is purely visual/layout.

### DIFF
--- a/GiniComponents/InternalPaymentSDK/GiniInternalPaymentSDK/Sources/GiniInternalPaymentSDK/PaymentReview/PaymentReviewObservableModel.swift
+++ b/GiniComponents/InternalPaymentSDK/GiniInternalPaymentSDK/Sources/GiniInternalPaymentSDK/PaymentReview/PaymentReviewObservableModel.swift
@@ -151,7 +151,7 @@ final class PaymentReviewObservableModel: ObservableObject {
     func paymentReviewPaymentInformationView(contentHeight: Binding<CGFloat>) -> PaymentReviewPaymentInformationView {
         PaymentReviewPaymentInformationView(viewModel: paymentInformationObservableModel,
                                             contentHeight: contentHeight,
-                                            showBanner: Binding( get: { self.showBanner }, set: { self.showBanner = $0 }),
+                                            showBanner: Binding(get: { self.showBanner }, set: { self.showBanner = $0 }),
                                             onBankSelectionTapped: { [weak self] in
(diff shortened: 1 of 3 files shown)


ValentinaIancu-Gini:
Nadya already raised a different pr which is fixing the issue based on a new design proposed. I will close this PR.

PR #1141, opened 2026-05-13
fix: Restructure green info bar layout above payment form fields

zladzeyka:
## Summary
[HEAL-387](https://ginis.atlassian.net/browse/HEAL-387)

Moves the green info bar from an overlay (which covered the top of the form fields) into the scroll content VStack so it sits **above** the form fields as designed.

## Changes
- Removed `.overlay(alignment: .top)` from `body` — banner no longer covers form content
- Banner is now the first element in the scroll content `VStack`, full-width with no horizontal padding
- Form fields are in an inner `VStack` with original horizontal/top padding
- `getHeight` now wraps the outer VStack so sheet height auto-adjusts when the banner appears/disappears
- Moved `.transition` and VoiceOver accessibility announcement to the banner's new location
- Removed unused `bannerYOffset = -8.0` constant

[HEAL-387]: https://ginis.atlassian.net/browse/HEAL-387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/GiniComponents/InternalPaymentSDK/GiniInternalPaymentSDK/Sources/GiniInternalPaymentSDK/PaymentReview/Views/PaymentInformation/PaymentReviewPaymentInformationView.swift
+++ b/GiniComponents/InternalPaymentSDK/GiniInternalPaymentSDK/Sources/GiniInternalPaymentSDK/PaymentReview/Views/PaymentInformation/PaymentReviewPaymentInformationView.swift
@@ -58,15 +58,6 @@ struct PaymentReviewPaymentInformationView: View {
     
     var body: some View {
         scrollView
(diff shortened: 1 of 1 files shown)


ValentinaIancu-Gini (approved):
@zladzeyka I tested this and everything works as expected. Very nice job!
I tested on multiple devices and it is working fine.

Tested on:

iPhone 15Pro - iOS 17.5
iPhone 17Pro- iOs26.4
iPhone 16 - iOS 18.6
iPhone X - iOS 16.7.15

PR #1169, opened 2026-06-02
fix(GiniCaptureSDK): Recover full EPC QR content when name field is null-byte padded

zladzeyka:
## Pull Request Description

[PP-2591](https://ginis.atlassian.net/browse/PP-2591)

### Root cause

Some QR code generators (e.g. Sparkasse) pad the EPC beneficiary name field with trailing spaces followed by null bytes (`0x00`). Apple's `AVMetadataMachineReadableCodeObject.stringValue` truncates the string at the first null byte, so the IBAN and all subsequent fields are silently lost. The SDK then sees a QR code starting with `BCD` but missing its IBAN, which previously caused either a crash (index out of range) or a silent format-not-valid rejection.

### Fix overview

**`Camera.swift`** — `qrCodeString(from:)` + `unpackByteModePayload(_:version:)`

When `stringValue` would truncate (detected by null bytes in the decoded content), the fix parses `CIQRCodeDescriptor.errorCorrectedPayload` directly. This is the raw QR data-codeword bit stream after error correction. Content bytes are packed at a 4-bit offset from byte boundaries, so a nibble-shift loop extracts them cleanly. `CIQRCodeDescriptor.symbolVersion` determines whether to read an 8-bit or 16-bit character count, avoiding ambiguity. Encoding is resolved by trying UTF-8 first (EPC character set "1"); Latin-1 is a guaranteed-non-nil fallback for character set "2" and also handles ISO 8859-1 umlauts correctly (UTF-8 decode fails on raw Latin-1 high bytes such as `0xDC` for Ü, triggering the fallback).

**`QRCodesExtractor.swift`**

- Trim trailing whitespace only on the beneficiary name (line 5) — Sparkasse pads it with 16 spaces before the null bytes.
- Strip internal spaces from the amount quantity in `normalize()` — Sparkasse produces `"EUR    10214.94"`.
- Add a line-10 unstructured reference fallback — Sparkasse leaves line 9 empty and uses line 10.

**`GiniQRCodeDocument.swift`**

- Strip null bytes defensively in `init` for any caller that passes a string with embedded `\0` characters (public API protection; Camera handles the truncation case before reaching here).
- Remove unused `epc06912LinesCount` property.

### Tests added

| Test | What it covers |
|---|---|
| `testEPC06912QRCodeMissingIBANDoesNotCrash` | Malformed QR (truncated at name) → format nil, no crash |
| `testEPC06912QRCodeWithNullBytePaddedNameExtractions` | Name with null-byte padding → all fields extracted correctly |
| `testEPC06912QRCodeSpacesAndNullPaddedNameWithAmountAndReferenceAtLine10` | Full Sparkasse-style QR: spaces + null padding, spaced amount, ref at line 10 |
| `testUnpackByteModePayloadNonByteModeReturnsNil` | Non-byte-mode payload → nil |
| `testUnpackByteModePayloadInsufficientBytesReturnsNil` | Declared count exceeds data → nil, no crash |
| `testUnpackByteModePayloadRoundtrip` | Manually constructed payload → correct bytes extracted |
| `testUnpackByteModePayloadISO88591UmlautFallsBackToLatin1` | `0xDC` (Ü in Latin-1) → UTF-8 fails → Latin-1 fallback gives correct result |
| `testUnpackByteModePayloadUTF8UmlautDecodesWithoutFallback` | `0xC3 0x9C` (Ü in UTF-8) → UTF-8 succeeds, no corruption |
| `testEPC06912QRCodeISO88591UmlautNameExtracts` | Umlaut in name (charset 2) passes through extraction unchanged |
| `testEPC06912QRCodeUTF8UmlautNameExtracts` | Umlaut in name (charset 1) passes through extraction unchanged |
| `testEPC06912QRCodeReferenceAtLine9StillExtractedWhenPresent` | Line-9 reference not shadowed by line-10 fallback |
| `testEPC06912QRCodeAmountWithoutSpacesNormalisesCorrectly` | Clean amounts unaffected by whitespace trim |
| `testEPC06912QRCodeBareCRSeparatorParsesCorrectly` | Bare `\r` (CR only) line endings parsed correctly |
| `testEPC06912QRCodeIBANWithSpacesIsAcceptedButStoredWithSpaces` | `IBANValidator` strips spaces to validate; spaces are preserved in `extractedParameters` |
| `testEPC06912QRCodeAmountWithCommaDecimalPreservedAsIs` | Comma-decimal amounts passed through as-is (documents current behaviour) |

### Test scenario (Sparkasse)

Scan a Sparkasse EPC QR whose name field is `"AUTOHAUS MARTIN WURST GMBH"` followed by 16 spaces and 29 null bytes.

- **Before:** `qrCodeFormat` is `nil` (truncated at null byte → IBAN lost → format rejected).
- **After:** All fields extracted correctly — IBAN, amount `"EUR    10214.94"`, and reference from line 10.

[PP-2591]: https://ginis.atlassian.net/browse/PP-2591

### DIFF
--- a/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
+++ b/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
@@ -96,18 +96,22 @@ public final class QRCodesExtractor {
         }
 
         if lines.indices.contains(5) && !lines[5].isEmpty {
-            parameters["paymentRecipient"] = lines[5]
+            // Sparkasse and similar generators pad the name with trailing spaces (and null bytes
+            // that Camera.swift strips before reaching here). Trim only this field.
+            parameters["paymentRecipient"] = lines[5].trimmingCharacters(in: .whitespaces)
         } else {
             parameters["paymentRecipient"] = ""
         }
 
+        // Field 9 (index 9) is the structured creditor reference; field 10 (index 10) is the
+        // unstructured remittance info. Some generators leave 9 empty and use 10 instead.
         if lines.indices.contains(9) && !lines[9].isEmpty {
             parameters["paymentReference"] = lines[9]
+        } else if lines.indices.contains(10) && !lines[10].isEmpty {
+            parameters["paymentReference"] = lines[10]
         } else {
             parameters["paymentReference"] = ""
         }
-
-        // if index out of range, return empty string
 
         if lines.indices.contains(6) && IBANValidator().isValid(iban: lines[6]) {
             parameters["iban"] = lines[6]
@@ -135,6 +139,7 @@ public final class QRCodesExtractor {
         if regexCurrency?.numberOfMatches(in: amount, options: [], range: NSRange(location: 0, length: length)) == 3 {
             let currency = String(amount[..<String.Index(utf16Offset: 3, in: amount)])
             let quantity = String(amount[String.Index(utf16Offset: 3, in: amount)...])
+                .trimmingCharacters(in: .whitespaces)
             return quantity + ":" + currency
         } else if let currency = currency {
             return amount + ":" + currency
--- a/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniQRCodeDocument.swift
+++ b/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniQRCodeDocument.swift
@@ -44,7 +44,6 @@ import GiniUtilites
     fileprivate let scannedString: String
     public lazy var extractedParameters: [String: String] = QRCodesExtractor
         .extractParameters(from: self.scannedString, withFormat: self.qrCodeFormat)
-    fileprivate let epc06912LinesCount = 12
     public lazy var qrCodeFormat: QRCodesFormat? = {
         if self.scannedString.starts(with: QRCodesFormat.giniQRCode.prefixURL) {
             return .giniQRCode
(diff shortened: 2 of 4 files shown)


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniQRCodeDocument.swift
```diff
@@ -54,11 +54,11 @@ import GiniUtilites
             return .eps4mobile
         } else if let lines = Optional(self.scannedString.splitlines),
                   lines.count > 0 && lines[0] == QRCodesFormat.epc06912.prefixURL {
-            if !(lines[2] == "1" || lines[2] == "2") {
+            if lines.indices.contains(2) && !(lines[2] == "1" || lines[2] == "2") {
                 print("WARNING: Character set \(lines[2]) is unknown. Expected version 1 or 2.")
```
qnaveed87: [Suggestion] Can we please replace `print` with `Log` ?

qnaveed87 (commented):
Thank you @zladzeyka I have minor suggestion

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniQRCodeDocument.swift
zladzeyka: Sure will do

qnaveed87 (approved):
Thank you @zladzeyka 

Looks good to me

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
```diff
@@ -96,19 +96,23 @@ public final class QRCodesExtractor {
...
+            parameters["paymentRecipient"] = lines[5].trimmingCharacters(in: .whitespaces)
         } else {
             parameters["paymentRecipient"] = ""
         }
 
+        // Field 9 (index 9) is the structured creditor reference; field 10 (index 10) is the
+        // unstructured remittance info. Some generators leave 9 empty and use 10 instead.
         if lines.indices.contains(9) && !lines[9].isEmpty {
             parameters["paymentReference"] = lines[9]
+        } else if lines.indices.contains(10) && !lines[10].isEmpty {
```
qnaveed87, line 110: [Suggestion] can we please add .trimmingCharacters(in: .whitespaces) for  paymentReference ?

qnaveed87 (approved):
Thank you @zladzeyka 

PR looks good to me I just left one minor suggestion

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
zladzeyka: @qnaveed87 why do we need it there?

PR #1182, opened 2026-06-22
Pp 2324 support new qr code types

abolfazlimahdi:
## Pull Request Description

Adds support for five new QR/barcode payment formats — SPC, SPD, PayBySquare, UPNQR, and HUB3 — extending GiniCaptureSDK beyond the existing EPC, EPS, Bezahl, and GiniQR formats.

[PP-2324]

**What changed and how:**

- **New formats detected and parsed:** `QRCodesFormat` enum extended with `.spc`, `.spd`, `.payBySquare`, `.upnqr`, and `.hub3`. Format detection in `GiniQRCodeDocument` uses prefix/first-line matching; parameter extraction in `QRCodesExtractor` implements format-specific field parsing for each.

- **PDF417 scanning enabled:** `Camera.swift` now registers both `.qr` and `.pdf417` metadata object types. HUB3 (Croatian payment slip) is distributed as a PDF417 barcode, not a QR code.

- **PayBySquare decoder:** Slovak PayBySquare encodes payment data as base32hex + LZ77-compressed binary. A pure Swift `LZMADecoder` (LZMA1 stream, no Apple compression framework dependency) and a `PayBySquareDecoder` were added to handle the full decode pipeline.

- **Validation:** `GiniCaptureDocumentValidator` updated with per-format validation rules. SPC, UPNQR, and HUB3 require a valid IBAN. SPD requires IBAN presence but skips strict validation (non-SEPA formats in the wild). PayBySquare validation is implied by a successful decode.

- **Unit tests:** `GiniQRCodeDocumentTests` covers format detection, field extraction, and validation for all five new formats using fixture strings.

## Notes for Reviewers

**Tested on:** iPhone 12 Pro, iOS 26.5

**Manual checks:**
- Format detection confirmed via unit tests for SPC, SPD, PayBySquare, UPNQR, and HUB3 fixture strings
- PayBySquare base32hex + LZMA1 decompression path tested against real-world encoded samples
- PDF417 scanning path tested in `Camera.swift` — both `.qr` and `.pdf417` are registered if the hardware supports them; no crash if `.pdf417` is unavailable
- Existing EPC, EPS, Bezahl, and GiniQR flows unaffected (only additive changes to the `switch` dispatch)

**Unit tests added:** `GiniCaptureSDKTests/GiniQRCodeDocumentTests.swift` — 262 lines covering all new formats

**Known limitation:** PayBySquare uses LZMA1 (not LZMA2); the pure Swift decoder handles the subset of LZMA1 streams produced by the PayBySquare spec. Apple's Compression framework only exposes LZMA2, which is why the custom decoder was necessary.

[PP-2324]: https://ginis.atlassian.net/browse/PP-2324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniQRCodeDocument.swift
```diff
@@ -51,18 +51,34 @@ import GiniUtilites
...
-                Log(message: "WARNING: Character set \(lines[2]) is unknown. Expected version 1 or 2.",
-                    event: "EPC QR code")
-            }
-
-            if lines.indices.contains(6) && IBANValidator().isValid(iban: lines[6]) {
-                return .epc06912
-            } else {
+        } else if self.scannedString.starts(with: QRCodesFormat.spd.prefixURL) {
+            return .spd
+        } else if let lines = Optional(self.scannedString.splitlines), !lines.isEmpty {
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: it is fixed

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquare/PayBySquareDecoder.swift
```diff
@@ -0,0 +1,153 @@
...
+        guard payloadLength > 4 else { return nil }
+
+        // 3. LZMA decompress bytes[4...]
+        guard let decompressed = lzmaDecompress(rawBody: Array(bytes[4...]),
+                                                payloadLength: payloadLength) else { return nil }
+
+        // 4. Skip 4-byte CRC32 prefix; decode remaining as UTF-8
+        guard decompressed.count > 4 else { return nil }
+        guard let text = String(bytes: Array(decompressed[4...]), encoding: .utf8) else { return nil }
+
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: it is fixed

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/GiniCaptureDocumentValidator.swift
```diff
@@ -109,6 +109,21 @@ fileprivate extension GiniCaptureDocumentValidator {
...
+        case .some(.spd):
+            // SPD IBANs include non-SEPA formats; require presence but skip strict IBAN validation
+            if document.extractedParameters["iban"] == nil {
+                throw DocumentValidationError.qrCodeFormatNotValid
+            }
+        case .some(.payBySquare):
+            // Successful decode is indicated by non-empty extractedParameters
+            if document.extractedParameters.isEmpty {
+                throw DocumentValidationError.qrCodeFormatNotValid
+            }
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: it is fixed

CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/GiniQRCodeDocumentTests.swift
```diff
@@ -484,3 +485,230 @@ final class GiniQRCodeDocumentTests: XCTestCase {
...
+    @Test func detectedAsPayBySquare() {
+        let doc = GiniQRCodeDocument(scannedString: detectionString)
+        #expect(doc.qrCodeFormat == .payBySquare)
+    }
+
+    @Test func nonBase32HexStringNotDetected() {
+        // Lowercase letters are outside the base32hex alphabet (0–9, A–V)
+        let doc = GiniQRCodeDocument(scannedString: "abcdef0123456789")
+        #expect(doc.qrCodeFormat != .payBySquare)
+    }
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: it is fixed

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/GiniCaptureDocumentValidator.swift
```diff
@@ -109,6 +109,20 @@ fileprivate extension GiniCaptureDocumentValidator {
             if document.extractedParameters[QRCodesExtractor.giniCodeUrlKey] == nil {
                 throw DocumentValidationError.qrCodeFormatNotValid
             }
+        case .some(.spc), .some(.upnqr), .some(.hub3):
+            guard let iban = document.extractedParameters["iban"],
+                  IBANValidator().isValid(iban: iban) else {
+                throw DocumentValidationError.qrCodeFormatNotValid
+            }
+        case .some(.spd):
```
zladzeyka, line 107: [Suggestion]: Unite 2 case, the code is identical there:
        `case .some(.spd), some(.payBySquare):`

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/LZMADecoder.swift
```diff
@@ -0,0 +1,376 @@
+//
+//  LZMADecoder.swift
+//  GiniCaptureSDK
```
qnaveed87: [Suggestion] Can ew please makes header consistent with other files like
//
//  LZMADecoder.swift
//
//  Copyright © 2026 Gini GmbH. All rights reserved
//

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquare/PayBySquareDecoder.swift
```diff
@@ -0,0 +1,119 @@
+//
+//  PayBySquareDecoder.swift
```
qnaveed87, line 2: Suggestion] Can ew please makes header consistent with other files like
//
// PayBySquareDecoder.swift
//
// Copyright © 2026 Gini GmbH. All rights reserved
//

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
```diff
@@ -0,0 +1,119 @@
...
+struct PayBySquarePayment {
+    let iban: String
+    let swift: String?
+    let amount: String
+    let currency: String
+    let payeeName: String
+    let paymentReference: String
+}
+
+final class PayBySquareDecoder {
```
qnaveed87: [Suggestion] Can we please add some description to this class as we did for `LZMADecoder`

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
```diff
@@ -0,0 +1,119 @@
...
+        let fields = text.components(separatedBy: "\t")
+        let banksCount = max(1, Int(fields.indices.contains(11) ? fields[11].trimmingCharacters(in: .whitespaces) : "") ?? 1)
+        // Extension fields are controlled by the PaymentOptions bitmask (field[2]):
+        //   bit 0 (standing order): +4 extension fields
+        //   bit 1 (direct debit):   +10 extension fields
+        let paymentOptions = Int(fields.indices.contains(2) ? fields[2] : "") ?? 0
+        let extFields = ((paymentOptions & 1) != 0 ? 4 : 0) + ((paymentOptions & 2) != 0 ? 10 : 0)
+        let beneficiaryIdx = 12 + banksCount * 2 + extFields
+
+        let iban = fields.indices.contains(12) ? fields[12] : ""
```
qnaveed87: [Request] Can we use an enum to mapped the fields like `private enum Field {     static let iban = 12 }
`

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/LZMADecoder.swift
```diff
@@ -0,0 +1,376 @@
...
+    // Uniform-probability bits (no probability model)
+    mutating func decodeDirectBits(numBits: Int) -> Int {
+        var result = 0
+        for _ in 0..<numBits {
+            range >>= 1
+            code &-= range                          // wrapping subtract
+            // If code underflowed (was < range), restore it and record a 1-bit
+            let underflow = Int(code >> 31)         // 1 if MSB set after subtract
+            if underflow != 0 { code &+= range }
+            result = (result << 1) | underflow
```
zladzeyka:
```suggestion
            result = (result << 1) | (1 &- underflow)
```
1 &- underflow uses Swift's wrapping subtract. Since underflow is always 0 or 1, wrapping never actually happens (1 - 0 = 1, 1 - 1 = 0), but &- matches the style of the surrounding &-= /
  &+= and avoids a compiler nag if you ever crank warnings.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
```diff
@@ -0,0 +1,119 @@
...
+    let amount: String
+    let currency: String
+    let payeeName: String
+    let paymentReference: String
+}
+
+final class PayBySquareDecoder {
+
+    static func looksLikePayBySquare(_ string: String) -> Bool {
+        let base32hexChars = CharacterSet(charactersIn: "0123456789ABCDEFGHIJKLMNOPQRSTUV")
```
zladzeyka: [Suggestion]: Please make sure that we accept here lower case as well `CharacterSet(charactersIn:
      "0123456789ABCDEFGHIJKLMNOPQRSTUVabcdefghijklmnopqrstuv")`
below in `decode` method you already normalize it to uppercase

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
```diff
@@ -0,0 +1,119 @@
...
+        //    [0] invoiceId      [1] paymentsCount  [2] paymentType  [3] amount
+        //    [4] currencyCode   [5] paymentDueDate [6] variableSymbol [7] constantSymbol
+        //    [8] specificSymbol [9] originatorRef  [10] paymentNote  [11] bankAccountsCount
+        //    [12] IBAN          [13] BIC           ...extensions...  [12+N*2+2] beneficiaryName
+        let fields = text.components(separatedBy: "\t")
+        let banksCount = max(1, Int(fields.indices.contains(11) ? fields[11].trimmingCharacters(in: .whitespaces) : "") ?? 1)
+        // Extension fields are controlled by the PaymentOptions bitmask (field[2]):
+        //   bit 0 (standing order): +4 extension fields
+        //   bit 1 (direct debit):   +10 extension fields
+        let paymentOptions = Int(fields.indices.contains(2) ? fields[2] : "") ?? 0
```
zladzeyka: [Clarification]: why we don't trim whitespaces here as we do for `bankCount` in line 57?

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
```diff
@@ -0,0 +1,119 @@
+//
+//  PayBySquareDecoder.swift
+//  GiniCaptureSDK
+//
+
+import Foundation
+
+struct PayBySquarePayment {
+    let iban: String
+    let swift: String?
```
zladzeyka:
```suggestion
    let bic: String?
```

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniQRCodeDocument.swift
```diff
@@ -51,23 +56,43 @@ import GiniUtilites
...
-                    event: "EPC QR code")
-            }
-
-            if lines.indices.contains(6) && IBANValidator().isValid(iban: lines[6]) {
-                return .epc06912
-            } else {
-                return nil
+        } else if self.scannedString.starts(with: QRCodesFormat.spd.prefixURL) {
+            return .spd
+        } else if let lines = Optional(self.scannedString.splitlines), !lines.isEmpty {
```
zladzeyka:
```suggestion
        } else if let lines = self.scannedString.splitlines, !lines.isEmpty {
```

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquare/LZMADecoder.swift
```diff
@@ -0,0 +1,376 @@
...
+// MARK: – Decoder state machine
+
+/// Holds the full mutable state of an in-progress LZMA1 decode and drives the
+/// decode loop. Splitting the loop body across `decodeLiteral`, `decodeNewMatch`,
+/// `decodeRep` and `copyMatch` keeps each step's branching small and isolated.
+private struct Machine {
+
+    // MARK: – Fixed bysquare parameters
+
+    private static let lc                  = 3
```
zladzeyka, line 52: [Suggestion]: For any reviewer unfamiliar with LZMA  very challenging to understand what are this lc, pb and other 1-3 names mean, please rename them

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
zladzeyka: [Suggestion]:Please use `enum` instead of `final class`

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
```diff
@@ -0,0 +1,119 @@
...
+                                  currency: currency,
+                                  payeeName: beneficiaryName,
+                                  paymentReference: reference)
+    }
+
+    // MARK: - Private
+
+    private static func buildReference(variableSymbol: String, paymentNote: String) -> String {
+        if !variableSymbol.isEmpty && !paymentNote.isEmpty {
+            return variableSymbol + " " + paymentNote
```
zladzeyka:
```suggestion
            return  "\(variableSymbol) \(paymentNote)"
```

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
```diff
@@ -0,0 +1,119 @@
...
+            if bitsStored >= 8 {
+                bitsStored -= 8
+                result.append(UInt8((accumulator >> bitsStored) & 0xFF))
+            }
+        }
+
+        return result
+    }
+
+
```
zladzeyka: [Suggestion]: Please remove blank lines

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
```diff
@@ -132,6 +160,144 @@ public final class QRCodesExtractor {
...
+            let parts = segment.split(separator: ":", maxSplits: 1).map(String.init)
+            guard parts.count == 2 else { continue }
+            let key = parts[0]
+            let value = parts[1]
+            switch key {
+            case "ACC": parameters["iban"] = value
+            case "AM":  parameters["amountToPay"] = value
+            case "CC":
+                if let amount = parameters["amountToPay"] {
+                    parameters["amountToPay"] = amount + ":" + value
```
zladzeyka:
```suggestion
                    parameters["amountToPay"] = "\(amount):\(value)"
```

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
```diff
@@ -132,6 +160,144 @@ public final class QRCodesExtractor {
...
+    // MARK: - Pay by Square (Slovak compressed QR)
+
+    class func extractParameters(fromPayBySquareString string: String) -> [String: String] {
+        guard let decoded = PayBySquareDecoder.decode(string) else { return [:] }
+        var parameters: [String: String] = [:]
+
+        parameters["iban"] = decoded.iban
+        parameters["paymentRecipient"] = decoded.payeeName
+        if !decoded.amount.isEmpty && !decoded.currency.isEmpty {
+            parameters["amountToPay"] = decoded.amount + ":" + decoded.currency
```
zladzeyka:
```suggestion
            parameters["amountToPay"] = "\(decoded.amount):\(decoded.currency)"
```

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
```diff
@@ -132,6 +160,144 @@ public final class QRCodesExtractor {
...
+
+    // MARK: - UPNQR (Slovenian UPN QR)
+
+    class func extractParameters(fromUPNQRCodeString string: String) -> [String: String] {
+        let lines = string.splitlines
+        var parameters: [String: String] = [:]
+
+        if lines.indices.contains(8), let cents = Int(lines[8]) {
+            let amount = String(format: "%.2f", Double(cents) / 100.0)
+            parameters["amountToPay"] = amount + ":EUR"
```
zladzeyka:
```suggestion
            parameters["amountToPay"] = "\(amount):EUR"
```

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
```diff
@@ -13,19 +13,47 @@ public enum QRCodesFormat {
...
+        static let bezahlPrefix      = "bank://"
+        // A Gini payment QR code is an HTTPS link to the Gini pay host. Assembled from
+        // its scheme and host parts so the detection marker is not a single hardcoded
+        // absolute-URI literal — it is a fixed format specifier, not a configurable endpoint.
+        static let giniQRCodeScheme  = "https"
+        static let giniQRCodeHost    = "pay.gini.net"
+        static let giniQRCodePrefix  = "\(giniQRCodeScheme)://\(giniQRCodeHost)/"
+        static let spcPrefix         = "SPC"
+        static let spdPrefix         = "SPD*"
+        static let payBySquarePrefix = ""
```
qnaveed87, line 58: [Clarification] Can you please tell why it does not have any prefix value unlike other QRCodes type and mention it in inline comments

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
zladzeyka:
```suggestion
    class func extractParameters(fromBezahlCodeString string: String) -> [String: String] {
```

CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/GiniQRCodeDocumentTests.swift
```diff
@@ -484,3 +485,264 @@ final class GiniQRCodeDocumentTests: XCTestCase {
                                                                    withConfig: giniConfiguration))
     }
 }
+
+// MARK: - SPC (Swiss Payment Code / QR-bill)
+
+@Suite("SPC QR Code")
+struct SPCQRCodeTests {
+
+    let config = GiniConfiguration()
```
zladzeyka: [Clarification]: how do we use config here? Please remove it if it's not used across the file.

zladzeyka (changes requested):
Thanks for shipping this 💙  solid work overall, especially the LZMA1 decoder from scratch and the Swift Testing adoption for new suites.👏🏻
 Left 15 inline comments.

Quick summary:
  Blocker:
  - LZMADecoder.decodeDirectBits bit sign inverted vs. the LZMA spec. Silent corruption for any match distance ≥ 128 bytes. Needs a fix + regression test with a large payload before merge.
  
  Should-fix in this round:
  - PayBySquareDecoder: lowercase detection, fields[2] trim asymmetry, banksCount bound.
  - QRCodesExtractor SPD parser: CC before AM field-order bug.
  - GiniQRCodeDocument:61: if let lines = Optional(...) — Kotlin .let{} mistranslation.
  - Validator: modernize Bezahl/EPC branch, merge .spd/.payBySquare cases.
  
  Stylistic (Swift idiom leaks worth flagging for the team): class func → static func, final class static-only → enum, string + → interpolation, .some(.x) → .x?, en-dash MARKs (macOS
  smart-dashes), swift: String? → bic.
  
  I'll do another round once the blocker is addressed. 🙏

qnaveed87 (commented):
Thank you @abolfazlimahdi and @MozhganPeivandianSharbaf for the impressive work.

I have left some minor comments please let me know if its not clear

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/LZMADecoder.swift
ValentinaIancu-Gini: @abolfazlimahdi and @MozhganPeivandianSharbaf we have a skill/instruction inside the repo that can help you with the style we have inside the ios repo.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquare/LZMADecoder.swift
```diff
@@ -0,0 +1,376 @@
...
+    // After new match: states < 7 → 7, states 7..11 → 10
+    private static let matchNextState:    [Int] = [7, 7, 7, 7, 7, 7, 7, 10, 10, 10, 10, 10]
+    // After rep (length > 1): states < 7 → 8, else → 11
+    private static let repNextState:      [Int] = [8, 8, 8, 8, 8, 8, 8, 11, 11, 11, 11, 11]
+    // After short rep (length == 1): states < 7 → 9, else → 11
+    private static let shortRepNextState: [Int] = [9, 9, 9, 9, 9, 9, 9, 11, 11, 11, 11, 11]
+
+    // MARK: – Mutable state
+
+    private var p = Probs()
```
ValentinaIancu-Gini: [Suggestion] Normally, we avoid using names with one, 2 3 letters. For better understanding, we should use a name that everyone can understand when reading the code. Can you please ask claude to renameame this :D ?

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/LZMADecoder.swift
ValentinaIancu-Gini: Also, this file is super long; let's try to split the logic into different files and maybe group them into a separate folder with a proper name for what the scope for this

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
```diff
@@ -35,16 +63,16 @@ public final class QRCodesExtractor {
...
+        case .some(.bezahl):      return extractParameters(fromBezhalCodeString: string)
+        case .some(.epc06912):    return extractParameters(fromEPC06912CodeString: string)
+        case .some(.eps4mobile):  return [epsCodeUrlKey: string]
+        case .some(.giniQRCode):  return [giniCodeUrlKey: string]
+        case .some(.spc):         return extractParameters(fromSPCCodeString: string)
+        case .some(.spd):         return extractParameters(fromSPDCodeString: string)
+        case .some(.payBySquare): return extractParameters(fromPayBySquareString: string)
+        case .some(.upnqr):       return extractParameters(fromUPNQRCodeString: string)
+        case .some(.hub3):        return extractParameters(fromHUB3CodeString: string)
+        case .none:               return [:]
```
ValentinaIancu-Gini: [Suggestion] Normally, each return line should go on a new line. Can you please fix this? Thank you!

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
```diff
@@ -132,6 +160,144 @@ public final class QRCodesExtractor {
...
+    }
+
+    private static func spcIBAN(from lines: [String]) -> String? {
+        guard lines.indices.contains(3), IBANValidator().isValid(iban: lines[3]) else { return nil }
+        return lines[3]
+    }
+
+    private static func spcAmount(from lines: [String]) -> String? {
+        guard let amount = nonEmptyLine(lines, at: 18) else { return nil }
+        let currency = nonEmptyLine(lines, at: 19) ?? "CHF"
```
ValentinaIancu-Gini, line 208: [Clarification] Is this the Swiss QR code type and that's why the default currency is CHF? Can we add a  // MARK: - Swiss qr code before ` private static func spcIBAN(from lines: [String]) -> String? {` ?

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniQRCodeDocument.swift
```diff
@@ -51,23 +56,43 @@ import GiniUtilites
...
+            case QRCodesFormat.epc06912.prefixURL:
+                return Self.epc06912Format(from: lines)
+            case QRCodesFormat.spc.prefixURL:
+                return Self.spcFormat(from: lines)
+            case QRCodesFormat.upnqr.prefixURL:
+                return .upnqr
+            case QRCodesFormat.hub3.prefixURL:
+                return .hub3
+            default:
+                return PayBySquareDecoder.looksLikePayBySquare(self.scannedString) ? .payBySquare : nil
```
ValentinaIancu-Gini, line 77: [Clarification] Is there any reason why the default case should be `payBySquare`?

CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/GiniQRCodeDocumentTests.swift
```diff
@@ -484,3 +485,264 @@ final class GiniQRCodeDocumentTests: XCTestCase {
                                                                    withConfig: giniConfiguration))
     }
 }
+
+// MARK: - SPC (Swiss Payment Code / QR-bill)
+
+@Suite("SPC QR Code")
+struct SPCQRCodeTests {
```
ValentinaIancu-Gini: [suggestion] This file is too long. Let's move all the new tests to separate files.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
```diff
@@ -132,6 +160,144 @@ public final class QRCodesExtractor {
...
+        return normalize(amount: amount, currency: currency)
+    }
+
+    private static func spcReference(from lines: [String]) -> String? {
+        guard lines.indices.contains(27), lines[27] != "NON" else { return nil }
```
ValentinaIancu-Gini, line 215: [Clarification] What is NON in this case?

ValentinaIancu-Gini (commented):
Thank you so much for creating this PR. I reviwed as well and left a couple of suggestions. Let's talk if there are unclear things.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/LZMADecoder.swift
MozhganPeivandianSharbaf: Fixed in `02f711ecf`. This turned out to be the real blocker — the decoded bit was inverted vs. the LZMA reference: it emitted `underflow` where the spec emits `1 - underflow`, silently corrupting any decode with a back-reference distance ≥ 128 bytes (the only path that reaches `decodeDirectBits`). Applied your `1 &- underflow` suggestion and added an `LZMADecoderTests` regression suite using an LZMA1 vector that forces a 256-byte match distance — verified it fails on the old code and passes on the fix. Thanks for catching this 🙏

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
MozhganPeivandianSharbaf: Fixed in `44647053f` — the charset now includes lowercase `a–v`, consistent with the uppercasing already done in `decode`.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
MozhganPeivandianSharbaf: Good catch — fixed in `44647053f`. `fields[2]` is now trimmed the same way as `fields[11]`.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
MozhganPeivandianSharbaf: Fixed in `44647053f` — renamed `swift` → `bic` in `PayBySquarePayment` and all usages.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniQRCodeDocument.swift
MozhganPeivandianSharbaf: Fixed in `44647053f`. Small note: `splitlines` returns a non-optional `[String]`, so `if let lines = self.scannedString.splitlines` wouldn't compile. I kept your intent (removing the pointless `Optional()` wrap) by moving it into the final `else`:

```swift
} else {
    let lines = self.scannedString.splitlines
    guard !lines.isEmpty else { return nil }
    switch lines[0] { ... }
}
```

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquare/LZMADecoder.swift
MozhganPeivandianSharbaf: `lc` / `lp` / `pb` are the canonical parameter names from the LZMA specification (Igor Pavlov's reference SDK) and are spelled out in the file header (lines 5–7: `lc = 3, lp = 0, pb = 2`). Renaming them away from the spec vocabulary would make cross-referencing the reference decoder harder, so I'd prefer to keep them and lean on the header doc. Happy to expand the inline comments if that helps readability — let me know if you feel strongly.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
MozhganPeivandianSharbaf: Fixed in `44647053f` — converted `final class PayBySquareDecoder` to an `enum` (it's a static-only type).

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
MozhganPeivandianSharbaf: Fixed in `44647053f` — now uses interpolation.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
MozhganPeivandianSharbaf: Fixed in `44647053f` — blank lines removed.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
MozhganPeivandianSharbaf: Fixed in `44647053f`, and I went a bit further: this parser is now field-order independent. `AM` and `CC` can appear in any order, and the previous code dropped the currency when `CC` came before `AM`. Amount and currency are now collected separately and combined (via interpolation) after the loop, with a regression test `combinesAmountAndCurrencyWhenCCPrecedesAM`.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
MozhganPeivandianSharbaf: Fixed in `44647053f` — now uses interpolation.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
MozhganPeivandianSharbaf: Fixed in `44647053f` — now uses interpolation.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
MozhganPeivandianSharbaf: Fixed in `44647053f` — corrected the `fromBezhalCodeString` typo to `fromBezahlCodeString` (declaration + call site).

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/GiniCaptureDocumentValidator.swift
MozhganPeivandianSharbaf: Fixed in `44647053f` — merged into `case .some(.spd), .some(.payBySquare):` with a shared body. I also modernized the Bezahl/EPC branch just above it: removed the dead `qrCodeFormat == nil` check (unreachable inside a `.some` case) and the redundant `isEmpty` / `== nil` conditions, replacing them with a single `guard let iban = ..., IBANValidator().isValid(iban: iban)`.

CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/GiniQRCodeDocumentTests.swift
MozhganPeivandianSharbaf: Fixed in `44647053f` — removed the unused `config` from the SPC, SPD, UPNQR and HUB3 suites. It's kept only in the Pay-by-Square suite, which actually calls `validate(_:withConfig:)`.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/LZMADecoder.swift
MozhganPeivandianSharbaf: Fixed in `363586148` — header is now consistent with the recent-file convention:

```
//
//  LZMADecoder.swift
//
//  Copyright © 2026 Gini GmbH. All rights reserved.
//
```

While here I also addressed @ValentinaIancu-Gini's two follow-ups on this thread: the file was split into a scoped `Core/Helpers/PayBySquare/` folder (`LZMADecoder.swift`, `LZMARangeDecoder.swift`, `PayBySquareDecoder.swift`) — `RangeDecoder` extracted into its own file, dropping `LZMADecoder.swift` from 378 → 288 lines — and all `///` declaration docs were converted to `/** … */` per `swift-documentation-style.md`.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquare/PayBySquareDecoder.swift
MozhganPeivandianSharbaf: Fixed in `363586148` — same banner applied:

```
//
//  PayBySquareDecoder.swift
//
//  Copyright © 2026 Gini GmbH. All rights reserved.
//
```

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
MozhganPeivandianSharbaf: Done in `363586148` — added a `/** … */` description on `PayBySquareDecoder` (and on `PayBySquarePayment`) explaining the decode pipeline: base32hex decode → parse the bysquare header → LZMA1 decompress → skip CRC32 → split the UTF-8 tab-separated fields. Used `/** */` rather than `///` to follow the repo's `swift-documentation-style.md`.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquareDecoder.swift
MozhganPeivandianSharbaf: Done in `363586148` — added a `private enum Field` for the tab-separated indices and a small `field(_:at:)` safe-access helper, so `decode` now reads `field(fields, at: Field.iban)`, `Field.amount`, etc. instead of raw `12`/`3`/… literals.

```swift
private enum Field {
    static let paymentType       = 2   // PaymentOptions bitmask
    static let amount            = 3
    static let currency          = 4
    static let variableSymbol    = 6
    static let paymentNote       = 10
    static let bankAccountsCount = 11
    static let iban              = 12
    static let bic               = 13
}
```

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
MozhganPeivandianSharbaf: Good question — clarified with an inline comment in `363586148`. Pay-by-Square carries no leading marker string (unlike the other formats, which start with a fixed specifier like `SPC`/`SPD*`/`HRVHUB30`). It's detected heuristically via `PayBySquareDecoder.looksLikePayBySquare` (base32hex alphabet + length ≥ 16), so the prefix is intentionally empty.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquare/LZMADecoder.swift
MozhganPeivandianSharbaf: Done in `4b2132a09` — renamed `p` → `probs` and `rc` → `rangeDecoder` for readability.

I intentionally kept `lc` / `lp` / `pb` (and `rep0…rep3`): those are the canonical LZMA-spec parameter names, documented in the file header (`lc = 3, lp = 0, pb = 2`). Renaming them away from the spec would make cross-referencing Igor Pavlov's reference decoder harder. Happy to also spell out `m` (the bit-tree node index) if you'd like.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
MozhganPeivandianSharbaf: Done in `4b2132a09` — each case in the `extractParameters(from:withFormat:)` switch now has its `return` on its own line.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
MozhganPeivandianSharbaf: Yes — this is the Swiss QR-bill (SPC) type, and CHF is its spec default currency. There's already a `// MARK: - SPC (Swiss Payment Code / QR-bill)` above `extractParameters(fromSPCCodeString:)` grouping these helpers, and I added an inline comment on the `?? "CHF"` fallback in `4b2132a09` to make the reasoning explicit.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniQRCodeDocument.swift
MozhganPeivandianSharbaf: Pay-by-Square is the only format with no leading marker string, so it can't be matched by a prefix `case`. It's the fallback — but only when `looksLikePayBySquare` (base32hex alphabet + length heuristic) confirms it; otherwise the format is treated as unknown (`nil`). Added a comment explaining this in `4b2132a09`.

CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/GiniQRCodeDocumentTests.swift
MozhganPeivandianSharbaf: Done in `4b2132a09` — moved the new suites into `Tests/GiniCaptureSDKTests/QRCodeFormats/`, one file per format (SPC, SPD, Pay-by-Square, LZMA, UPNQR, HUB3). The original `GiniQRCodeDocumentTests` keeps the existing `XCTestCase`. All 30 tests still pass.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
MozhganPeivandianSharbaf: `NON` is the Swiss QR-bill reference-type token meaning "no reference" (as opposed to `QRR` / `SCOR`), so there's nothing to extract in that case. Added an inline comment in `4b2132a09`.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/LZMADecoder.swift
MozhganPeivandianSharbaf: Done — applied `swift-documentation-style.md`: all `///` declaration docs were converted to `/** … */` (in `363586148`). Thanks for pointing to the instruction! 🙏

MozhganPeivandianSharbaf: Done — split into a scoped `Core/Helpers/PayBySquare/` folder (`LZMADecoder`, `LZMARangeDecoder`, `PayBySquareDecoder`) in `363586148`; `LZMADecoder.swift` is down from 378 → 288 lines.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquare/LZMADecoder.swift
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        // Slots 14+: (numDirBits - 4) direct bits + 4 align bits
+        let distBase = (2 | (posSlot & 1))
+        let directBits = rangeDecoder.decodeDirectBits(numBits: numDirBits - Self.numAlignBits)
+        var dist = UInt32(distBase) << UInt32(numDirBits)
+        dist |= UInt32(directBits) << UInt32(Self.numAlignBits)
+        dist |= UInt32(rangeDecoder.decodeReverseBitTree(probs: &probs.alignProbs,
+                                                         offset: -1,
+                                                         numBits: Self.numAlignBits))
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: This is a false positive — `offset: -1` is intentional and correct here.

`alignProbs` is a dedicated 16-entry table used **only** by this align tree. With `offset: -1`, tree node `m` (1…15) maps to slot `m-1` (0…14) — a consistent relabeling: every node reads/writes the same slot, the array is uniformly initialised to 1024, and no other code touches it. The decoded value is mathematically identical to the conventional `1…15` layout, there's no index collision, and no out-of-bounds (max index is 14 < 16).

It's also proven empirically: `LZMADecoderTests.decodesLongDistanceMatchCorrectly` decodes a payload with a **256-byte match distance** (`posSlot 16`, i.e. ≥ 14), which goes through exactly this align-tree path, and reproduces all 512 bytes exactly. So distance decoding for `posSlot >= 14` is verified correct. No change needed.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquare/PayBySquareDecoder.swift
```diff
@@ -0,0 +1,163 @@
...
+        // 3. LZMA1 decompress bytes[4...]
+        // Apple's COMPRESSION_LZMA only handles XZ/LZMA2; bysquare uses raw LZMA1.
+        guard let decompressed = LZMADecoder.decode(input: Array(bytes[4...]),
+                                                    outputLength: payloadLength) else { return nil }
+
+        // 4. Skip 4-byte CRC32 prefix; decode remaining as UTF-8
+        // Note: CRC32 validation was attempted but rejected valid codes from the field — the bysquare
+        // spec's exact CRC32 variant and byte layout need a verified test vector before re-enabling.
+        guard decompressed.count > 4 else { return nil }
+        guard let text = String(bytes: Array(decompressed[4...]), encoding: .utf8) else { return nil }
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: Valid observation, but this is a deliberate, documented deferral (see the note at the skip site). CRC32 validation was attempted and rejected valid field codes — the exact bysquare CRC32 variant/byte order needs a verified test vector before it can be enabled without breaking real payments. Tracked as a follow-up; the field values are still validated downstream (non-empty IBAN, etc.).

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniQRCodeDocument.swift
```diff
@@ -45,29 +50,52 @@ import GiniUtilites
...
+                // be matched by a prefix case. It's the fallback, but only when the
+                // base32hex heuristic confirms it; otherwise the format is unknown (nil).
+                return PayBySquareDecoder.looksLikePayBySquare(self.scannedString) ? .payBySquare : nil
+            }
         }
     }()
 
+    /// EPC069-12 (Stuzza/GiroCode): valid only when line 6 holds a valid IBAN.
+    /// Line 2 carries the character-set version (1 or 2); anything else is logged.
+    private static func epc06912Format(from lines: [String]) -> QRCodesFormat? {
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: Fixed in `bc27f0f18` — converted the `epc06912Format` declaration doc to `/** … */` per `swift-documentation-style.md`. Thanks, this was missed in the earlier house-style sweep.

CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/GiniQRCodeDocumentTests.swift
```diff
@@ -6,6 +6,7 @@
 //  Copyright © 2017 Gini GmbH. All rights reserved.
 //
 
+import Testing
 import XCTest
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: Fixed in `bc27f0f18` — removed the now-unused `import Testing`; this file is pure `XCTestCase` after the Swift Testing suites were split into `QRCodeFormats/`.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Models/GiniQRCodeDocument.swift
```diff
@@ -45,29 +50,52 @@ import GiniUtilites
...
+        if lines.indices.contains(2) && !(lines[2] == "1" || lines[2] == "2") {
+            Log(message: "WARNING: Character set \(lines[2]) is unknown. Expected version 1 or 2.",
+                event: "EPC QR code")
+        }
+        guard lines.indices.contains(6), IBANValidator().isValid(iban: lines[6]) else { return nil }
+        return .epc06912
+    }
+
+    /// SPC / Swiss QR-bill: valid only when line 3 holds a valid IBAN.
+    private static func spcFormat(from lines: [String]) -> QRCodesFormat? {
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: Fixed in `bc27f0f18` — converted the `spcFormat` declaration doc to `/** … */` per the repo doc style.

qnaveed87 (approved):
Thanks you @MozhganPeivandianSharbaf 

Looks good to me

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquare/PayBySquareDecoder.swift
```diff
@@ -0,0 +1,163 @@
...
+        //    Then two extension-presence flags, any present extension blocks, then:
+        //    [12 + N*2 + 2 (+ extension fields)] beneficiaryName
+        let fields = text.components(separatedBy: "\t")
+        let rawBanksCount = max(1, Int(field(fields, at: Field.bankAccountsCount).trimmingCharacters(in: .whitespaces)) ?? 1)
+        // Bound by field count so a malformed payload can't push beneficiaryIdx far past the array.
+        let banksCount = min(rawBanksCount, fields.count)
+        // PaymentOptions bitmask (Field.paymentType): paymentorder = 1, standingorder = 2,
+        // directdebit = 4. A present standing order adds 4 fields; a present direct debit adds 10.
+        let paymentOptions = Int(field(fields, at: Field.paymentType).trimmingCharacters(in: .whitespaces)) ?? 0
+        let extensionFields = ((paymentOptions & 2) != 0 ? 4 : 0)   // standing order block
```
zladzeyka, line 105: [Suggestion]: Both test vectors use `paymentOptions=1`, so this branch always evaluates to 0 in tests. The whole reason `beneficiaryIdx` is nontrivial — skipping standing-order (4 fields) or direct-debit (10 fields) blocks — has no coverage. If the spec's block sizes or ordering are off, standing-order and direct-debit QRs will silently return the wrong `paymentRecipient`.

Can you add a test with `paymentOptions & 2` (standing order) or `& 4` (direct debit)? A synthetic vector like `beneficiaryVector` but with the extension flag set would be enough.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
```diff
@@ -132,6 +175,165 @@ public final class QRCodesExtractor {
...
+        if lines.indices.contains(6) && !lines[6].isEmpty {
+            parameters["paymentRecipient"] = lines[6]
+        }
+        if lines.indices.contains(9) {
+            parameters["iban"] = lines[9]
+        }
+        // Croatian "poziv na broj" is the model (line 10, e.g. "HR01") followed by the
+        // reference number (line 11); the current parser dropped the model. Combine them,
+        // falling back to whichever part is present.
+        // NOTE: assumes the backend expects the combined poziv na broj — verify against a
```
zladzeyka: [Clarification]: The `NOTE:` says "verify against a real HUB3 payload before relying on the exact format" — that verification should happen before merge, not after. Two questions:

1. Does the backend expect `"HR00 1234567890"` (space-joined, as implemented) or `"HR001234567890"` (concatenated)?
2. Is there a real HUB3 payload available to test against?

Silently sending the wrong reference format for Croatian payments would be hard to detect from telemetry.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/GiniCaptureDocumentValidator.swift
```diff
@@ -109,6 +107,17 @@ fileprivate extension GiniCaptureDocumentValidator {
             if document.extractedParameters[QRCodesExtractor.giniCodeUrlKey] == nil {
                 throw DocumentValidationError.qrCodeFormatNotValid
             }
+        case .some(.spc), .some(.upnqr), .some(.hub3):
+            guard let iban = document.extractedParameters["iban"],
+                  IBANValidator().isValid(iban: iban) else {
+                throw DocumentValidationError.qrCodeFormatNotValid
+            }
+        case .some(.spd), .some(.payBySquare):
```
zladzeyka: [Suggestion]: The `.payBySquare` validation branch only checks `!iban.isEmpty`, which is looser than `.spc`, `.upnqr`, and `.hub3` (all check-digit validate). Since Pay-by-Square is Slovak-only and always SEPA, there's no reason for the softer check — and it's the difference between "corrupt scan gets rejected" and "corrupt scan produces a wrong-IBAN payment prompt."

Suggested split:

```swift
case .some(.payBySquare):
    guard let iban = document.extractedParameters["iban"],
          IBANValidator().isValid(iban: iban) else {
        throw DocumentValidationError.qrCodeFormatNotValid
    }
case .some(.spd):
    guard let iban = document.extractedParameters["iban"], !iban.isEmpty else {
        throw DocumentValidationError.qrCodeFormatNotValid
    }
```

Existing tests (`officialTestVectorPassesValidation`, `beneficiaryVectorDecodesIBANAndAmount`) already use valid SK IBANs and will continue to pass.

Separately, could you file a ticket to identify the bysquare CRC32 variant using the official test vector already in the suite? Skipping CRC32 opens a narrow but real hole — a garbled LZMA1 stream that decompresses into valid UTF-8 with a plausible IBAN would be accepted. That's the deeper fix but doesn't need to block this PR.

zladzeyka: And could you please take about Sonar Cloud issue?

MozhganPeivandianSharbaf: Done. Split `.payBySquare` out and it now goes through full `IBANValidator().isValid()` — folded into the existing strict `.spc/.upnqr/.hub3` case so we do not reintroduce the duplication from the earlier round; `.spd` keeps the non-empty-only check.

On CRC32: rather than deferring to a ticket, I identified the variant and enabled validation. Using the official test vector I confirmed it is standard CRC-32 (ISO-HDLC / zlib, poly `0xEDB88320`) computed over the payload bytes and stored **little-endian** (`zlib.crc32(data) == 0x61A4DF60`, matching the prefix) — the earlier attempt most likely failed on endianness/byte-range. `PayBySquareDecoder` now rejects any payload whose checksum does not match. I also regenerated the `beneficiaryVector` fixture (it carried a zeroed CRC) and added negative tests. (commit 7d8134e)

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/QRCodesExtractor.swift
MozhganPeivandianSharbaf: Verified the expected format: the HUB3 payment reference is **line 11 (call number) only** — the model line (10) is not part of the reference sent to the backend, so it is neither space-joined nor concatenated. Updated the iOS extractor to use line 11 verbatim, removed the unverified `NOTE`, and updated the HUB3 tests accordingly. (commit 2843641)

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Helpers/PayBySquare/PayBySquareDecoder.swift
MozhganPeivandianSharbaf: Added coverage in a new `Pay by Square field mapping` suite: tests for `paymentOptions` = standing order (`&2`), direct debit (`&4`), both (6), and multi-account. To make this testable without a full base32hex+LZMA round-trip I extracted the field→payment mapping into `PayBySquareDecoder.makePayment(fromFields:)`, and the tests lay the fields out **positionally** (independent of the decoder’s index formula), so a regression in the `beneficiaryIdx` arithmetic fails the test. (commit 2843641)

Note: these tests share the same block-size constants the decoder assumes (2 presence flags, +4 standing order, +10 direct debit), so they lock in the current behaviour but cannot independently verify those constants — that would need a real captured standing-order / direct-debit QR. Flagging that as a follow-up rather than guessing.

zladzeyka (approved):
Thanks you @abolfazlimahdi @MozhganPeivandianSharbaf

zladzeyka (approved):

PR #1211, opened 2026-07-24
fix(GiniCaptureSDK): Set take-photo as preferred action on unsupporte…

qnaveed87:
## Pull Request Description

Updates the GiniCaptureSDK camera flow so that, when an unsupported QR code is detected, the alert defaults to the “Take photo of document” action to better guide the user.

## Notes for Reviewers

- Introduces a dedicated UIAlertAction instance for “Take photo of document”.
- Sets that action as alert.preferredAction in the unsupported QR alert.


ValentinaIancu-Gini (approved):
Looks good to me! @qnaveed87

PR #1212, opened 2026-07-27
Pp 3238 update label after qr optout

qnaveed87:
## Pull Request Description

[PP-3238](https://ginis.atlassian.net/browse/PP-3238)

This PR updates CameraViewController (CaptureSDK) to keep the camera screen’s navigation title consistent after a user opts out of QR scanning via “Take photo of document”, especially when the device rotates.

## Notes for Reviewers

- Adds an “opt-out” check inside configureTitle() intended to preserve the “Scan invoice” title after QR opt-out.
Introduces showOnlyInvoiceTitles() and calls it when the user selects “Take photo of document” from the unsupported-QR alert.

[PP-3238]: https://ginis.atlassian.net/browse/PP-3238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/Camera/CameraViewController.swift
+++ b/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/Camera/CameraViewController.swift
@@ -143,8 +143,9 @@ final class CameraViewController: UIViewController {
 
     fileprivate func configureTitle() {
         let device = UIDevice.current
-        // True once the user has opted out of QR scanning via "Take photo of document".
-        let didOptOutOfQRScanning = cameraPane.cameraTitleLabel?.text == Strings.onlyInvoice
+        // The camera UI already shows "Scan invoice" — from initial config with QR disabled,
+        // or after the user tapped "Take photo of document".
+        let cameraTitleAlreadyReadsInvoice = cameraPane.cameraTitleLabel?.text == Strings.onlyInvoice
             || title == Strings.onlyInvoice
 
         if device.isIphone && device.isPortrait {
@@ -152,8 +153,8 @@ final class CameraViewController: UIViewController {
             return
         }
 
-        // Keep "Scan invoice" title across rotation after the user opted out of QR scanning.
-        if didOptOutOfQRScanning {
+        // Preserve the "Scan invoice" title on rotation after the opt-out.
+        if cameraTitleAlreadyReadsInvoice {
             title = Strings.onlyInvoice
             return
         }
--- a/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/Camera/CameraViewController.swift
+++ b/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/Camera/CameraViewController.swift
@@ -722,6 +722,8 @@ final class CameraViewController: UIViewController {
         cameraPreviewViewController.cameraFrameView.isHidden = false
         isQRScanFlowActive = false
         detectedQRCodeDocument = nil
+        cameraPane.cameraTitleLabel?.text = Strings.onlyInvoice
+        cameraPaneHorizontal?.cameraTitleLabel?.text = Strings.onlyInvoice
     }
 
     private func isAccessibilityLargeTextEnabled() -> Bool {
--- a/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/Camera/CameraViewController.swift
+++ b/CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/Camera/CameraViewController.swift
@@ -726,7 +726,7 @@ final class CameraViewController: UIViewController {
         detectedQRCodeDocument = nil
     }
 
-    private func handleTakePhotoOfDocument() {
+    func handleTakePhotoOfDocument() {
         // QR detection stays paused — resuming here would immediately re-trigger the alert
         cameraPreviewViewController.cameraFrameView.isHidden = false
         isQRScanFlowActive = false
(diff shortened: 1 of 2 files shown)


CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/Camera/CameraViewController.swift
```diff
@@ -143,12 +143,21 @@ final class CameraViewController: UIViewController {
...
         if device.isIphone && device.isPortrait {
             title = giniConfiguration.onlyQRCodeScanningEnabled ? Strings.onlyQr : Strings.cameraTitle
             return
         }
 
+        // Keep "Scan invoice" title across rotation after the user opted out of QR scanning.
+        if didOptOutOfQRScanning {
+            title = Strings.onlyInvoice
+            return
+        }
```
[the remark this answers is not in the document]
qnaveed87: Portrait "Scan" title is by design (bottom label shows "Scan invoice"). IT does not change visible output l- No changed needed

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/Camera/CameraViewController.swift
```diff
@@ -143,12 +143,21 @@ final class CameraViewController: UIViewController {
 
     fileprivate func configureTitle() {
         let device = UIDevice.current
+        // True once the user has opted out of QR scanning via "Take photo of document".
+        let didOptOutOfQRScanning = cameraPane.cameraTitleLabel?.text == Strings.onlyInvoice
```
zladzeyka: [Request]:  Suggest to implement explicit bool:
 private var didOptOutOfQRScanning = false
  // set in handleTakePhotoOfDocument(), read in configureTitle()

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/Camera/CameraViewController.swift
```diff
@@ -143,12 +143,21 @@ final class CameraViewController: UIViewController {
...
+            || title == Strings.onlyInvoice
 
         if device.isIphone && device.isPortrait {
             title = giniConfiguration.onlyQRCodeScanningEnabled ? Strings.onlyQr : Strings.cameraTitle
```
zladzeyka, line 152: [Clarification]: iPhone portrait is it intended? The if device.isIphone && device.isPortrait block returns before the new opt-out check, so on iPhone portrait the nav title becomes
  cameraTitle, not onlyInvoice, even after opt-out.

zladzeyka (changes requested):
Thank you @qnaveed87 for the fix. Please find my review notes. As additional suggestion I would recommend to add a test covering opt-out → rotate → title preserved.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/Camera/CameraViewController.swift
qnaveed87: Thanks for the suggestion! We already have a few flags in this flow, so I'd rather not introduce another one if we can avoid it. If you're okay with it, I'd prefer to stick with the current approach.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/Camera/CameraViewController.swift
qnaveed87: Yes, that's intentional. On iPhone in portrait we always return early and keep the navigation title as "Scan". After opting out of QR scanning, the bottom pane already shows "Scan invoice", so we don't repeat it in the navigation bar. The opt-out check is only applied on iPad and in landscape, where that bottom label isn't available.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/Camera/CameraViewController.swift
zladzeyka: Fair point on flag proliferation. Could you rename to something like `paneAlreadyShowsInvoiceTitle` and 
adjust the comment? Keeps the mechanism, just makes the check honest about what it reads.

CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/CameraPreviewViewControllerTests.swift
```diff
@@ -162,6 +162,75 @@ final class CameraPreviewViewControllerTests: XCTestCase {
...
+        giniConfiguration.qrCodeScanningEnabled = true
+        let viewController = CameraViewController(giniConfiguration: giniConfiguration,
+                                                  viewModel: CameraButtonsViewModel())
+        viewController.loadViewIfNeeded()
+
+        let onlyInvoiceLabel = NSLocalizedString("ginicapture.camera.infoLabel.only.invoice",
+                                                 bundle: giniCaptureBundle(),
+                                                 comment: "")
+
+        // Simulate the post-"Take photo of document" state (handleTakePhotoOfDocument is private).
```
zladzeyka: [Suggestion]: If we don't call `handleTakePhotoOfDocument` then we don't actually test the code above. We can make `handleTakePhotoOfDocument` internal or keep it private and inside call internal helper.

zladzeyka (changes requested):

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/Camera/Camera/CameraViewController.swift
qnaveed87: Done

CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/CameraPreviewViewControllerTests.swift
qnaveed87: Done

zladzeyka (approved):
LGTM

PR #1230, opened 2026-08-04
fix(GiniCaptureSDK): Preserve camera frame color across rotation

ValentinaIancu-Gini:
## Pull Request Description

Camera frame corners revert from red to white when device rotates while invalid-QR feedback is active.

[PP-3305]

`updateFrameOrientation` rebuilt the frame image from the pristine untinted `cameraFocus` asset on every rotation, discarding whatever tint (e.g. red `error3` for invalid QR) was previously applied via `changeCameraFrameColor(to:)`. Fix stores the currently applied color and reapplies it after the orientation rebuild.

## Notes for Reviewers

- Root cause diagnosis: `specs/PP-3305-bug.md`
- Regression test added: `CameraPreviewViewControllerTests.testCameraFrameKeepsColorAfterOrientationUpdate` — pixel-scans the frame image for red after simulating a rotation via `updatePreviewViewOrientation()`. Fails before the fix, passes after.
- `GiniCaptureSDK` full unit test suite passes; scheme builds clean.
- Manual QA still needed: real-device rotation with invalid QR + unsupported-QR alert on screen (iPhone + iPad).

[PP-3305]: https://ginis.atlassian.net/browse/PP-3305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/CameraPreviewViewControllerTests.swift
+++ b/CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/CameraPreviewViewControllerTests.swift
@@ -103,7 +103,9 @@ final class CameraPreviewViewControllerTests: XCTestCase {
                       "frame image should stay red after an orientation update")
     }
 
-    /// Scans the image for at least one mostly-opaque pixel whose color matches `color`.
+    /**
+     Scans the image for at least one mostly-opaque pixel whose color matches `color`.
+     */
     private func hasPixel(matching color: UIColor, in image: UIImage?) -> Bool {
         guard let cgImage = image?.cgImage else { return false }
         let width = cgImage.width
--- a/CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/CameraPreviewViewControllerTests.swift
+++ b/CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/CameraPreviewViewControllerTests.swift
@@ -85,6 +85,62 @@ final class CameraPreviewViewControllerTests: XCTestCase {
         XCTAssertNotEqual(defaultFlashState, camera.isFlashOn, "camera flash state should change it after toggle it")
     }
 
+    // MARK: - Camera frame color across rotation (PP-3305)
+
+    func testCameraFrameKeepsColorAfterOrientationUpdate() {
+        _ = cameraPreviewViewController.view
+        XCTAssertNotNil(UIImageNamedPreferred(named: "cameraFocus"),
+                        "cameraFocus asset must be available for this test to be meaningful")
+
+        cameraPreviewViewController.changeCameraFrameColor(to: .red)
+        XCTAssertTrue(hasPixel(matching: .red, in: cameraPreviewViewController.cameraFrameView.image),
+                      "frame image should be tinted red after changeCameraFrameColor")
+
+        // What viewWillTransition triggers on device rotation — it rebuilds the frame image.
+        cameraPreviewViewController.updatePreviewViewOrientation()
+
+        XCTAssertTrue(hasPixel(matching: .red, in: cameraPreviewViewController.cameraFrameView.image),
+                      "frame image should stay red after an orientation update")
+    }
+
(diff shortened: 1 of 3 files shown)


CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/CameraPreviewViewControllerTests.swift
```diff
@@ -85,6 +85,64 @@ final class CameraPreviewViewControllerTests: XCTestCase {
...
+        guard let context = CGContext(data: &data,
+                                      width: width,
+                                      height: height,
+                                      bitsPerComponent: 8,
+                                      bytesPerRow: width * 4,
+                                      space: colorSpace,
+                                      bitmapInfo: CGImageAlphaInfo.premultipliedLast.rawValue) else {
+            return false
+        }
+        context.draw(cgImage, in: CGRect(x: 0, y: 0, width: width, height: height))
```
abolfazlimahdi, line 124: **Should-fix: undefined behavior — `&data` outlives the call it's valid for.**

Passing `&data` to `CGContext(data:)` is an inout-to-pointer conversion that is only guaranteed valid for the duration of that initializer call. The context retains the pointer and writes through it later, during `context.draw(...)` — formally UB, even though it happens to work today.

Wrap the context creation, the draw, and the pixel scan in `data.withUnsafeMutableBytes`:

```swift
let found = data.withUnsafeMutableBytes { buffer -> Bool in
    guard let context = CGContext(data: buffer.baseAddress,
                                  width: width,
                                  height: height,
                                  bitsPerComponent: 8,
                                  bytesPerRow: width * 4,
                                  space: colorSpace,
                                  bitmapInfo: CGImageAlphaInfo.premultipliedLast.rawValue) else {
        return false
    }
    context.draw(cgImage, in: CGRect(x: 0, y: 0, width: width, height: height))
    // ... pixel scan over buffer ...
}
```

CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/CameraPreviewViewControllerTests.swift
```diff
@@ -85,6 +85,64 @@ final class CameraPreviewViewControllerTests: XCTestCase {
         XCTAssertNotEqual(defaultFlashState, camera.isFlashOn, "camera flash state should change it after toggle it")
     }
 
+    // MARK: - Camera frame color across rotation (PP-3305)
+
+    func testCameraFrameKeepsColorAfterOrientationUpdate() {
```
abolfazlimahdi, line 90: **Should-fix: new test uses XCTest — repo rule mandates Swift Testing.**

Per `.claude/rules/mandatory-rules.md` / CLAUDE.md, new tests MUST use Swift Testing (`@Suite`, `@Test`, `#expect`); XCTest is reserved for XCUITest and Objective-C suites. Swift Testing coexists with XCTest in the same target, so this could live as a small `@Suite` in this file:

```swift
@Suite("Camera frame color across rotation (PP-3305)")
struct CameraFrameColorRotationTests {
    @Test func frameKeepsColorAfterOrientationUpdate() { ... #expect(...) }
}
```

Matching the surrounding XCTest class is a defensible pragmatic call — but since it deviates from a MUST rule, it needs an explicit decision rather than a silent exception. Fine to keep as-is if we agree here in the thread.

abolfazlimahdi (changes requested):
I tested the fix on a real device and it is working fine — the frame keeps its red tint across rotation with the invalid-QR alert active. ✅

The fix itself resolves PP-3305 correctly (verified the tint re-application is lossless since `tintedImageWithColor` renders via `.alwaysTemplate`, and all frame-color state transitions in `CameraViewController` go through `changeCameraFrameColor`/`changeQRFrameColor`, so the stored color never goes stale).

Requesting changes only for the two test-quality items noted inline.

abolfazlimahdi (approved):
Looks good to me, thank you!

PR #1231, opened 2026-08-05
ai: Add gini-review PR review skill

MozhganPeivandianSharbaf:
## Pull Request Description

Adds `/gini-review`, a Claude Code skill that runs a complete PR review of this repository — resolving the PR from the current branch, a PR number or a Jira ticket key, checking every changed file against the ticket's acceptance criteria and this repo's iOS conventions, then reporting a coverage ledger plus triaged findings. It exists so reviews start from a consistent, repo-aware baseline instead of each reviewer re-deriving the local rules, and so the thorough pass is already done by the time a human opens the diff.

N/A 

The skill is split in two, mirroring the structure already in place in `gini-mobile-android`:

- **`pr-review/`** — the platform-neutral review engine: the procedure, the review dimensions, the confidence filter, the report template, comment style and Jira ticket handling. This is **byte-identical with the Android copy** at the same path and must not be edited here — Android owns the mirror through `.github/mirrored-skills.txt` and its `shared-skills.*` workflows.
- **`SKILL.md` + `references/ios-*.md`** — the iOS platform layer: Swift visibility and the seven public release repos, the `Package.swift` / `Package-release.swift` two-manifest rule, release mechanics, the module-ripple table, `CLAUDE.md`'s MUST-level architecture, style and test rules, and the do-not-flag list that keeps the output readable.

The skill never casts an approve or request-changes verdict, never modifies a file, and never runs builds or tests. Nothing reaches GitHub without an explicit confirmation.

## Notes for Reviewers

- **Skill-only change.** Everything is under `.claude/skills/gini-review/`. No SDK, example app or CI file is touched, so no tests apply and no module is affected.
- **Review `pr-review/` as a copy, not as new content** — confirm it matches Android rather than reading 969 lines line by line:

  ```bash
  diff -r <path-to-gini-mobile-android>/.claude/skills/gini-review/pr-review \
          .claude/skills/gini-review/pr-review && echo identical
  ```

  It is nested inside `gini-review/` rather than sitting at the top level on purpose: keeping the path identical to Android's is what lets the mirror check diff the two repos.
- **The iOS layer is where review effort is best spent.** It was derived from `AGENTS.md`, `CLAUDE.md` (including the "MyApp Standards" MUSTs), `.swiftlint.yml`, the `*.check.yml` workflows, `RELEASE-ORDER.md` and the per-module manifests. Three load-bearing claims worth a second pair of eyes:
  - **SwiftLint is a warning-only build phase** on the BankSDK and HealthSDK example projects — invoked without `--strict`, with every `error:` threshold commented out in `.swiftlint.yml` — and not a CI gate. The do-not-flag list depends on this being accurate.
  - **A `GiniUtilites`-only PR triggers no dependent SDK's checks.** `utilites.check.yml` matches only its own path, so on those PRs the human reviewer is the only gate on downstream breakage.
  - **Three points where `CLAUDE.md` has drifted from the code** are documented rather than silently resolved: the `UIColor.Gini*` colour mandate vs the newer `GiniColorScheme` (~25 files), "new tests MUST use Swift Testing" vs XCTest still being the majority by file count, and fixtures described as JSON when several modules ship `.pdf`, `.jpg`, `.txt` and `.png`. If any should be resolved the other way, the checklist is the place to say so.
- **Not yet verified:** whether the codebase actually honours `CLAUDE.md`'s *ViewModels MUST NOT import UIKit* and *view controllers hold no business logic*. Both are cited as blocking-grade rules; if they turn out to be aspirational, they belong in the drift list too.
- **To try it:** run `/gini-review <pr-number>` against any open PR. Output is terminal-first and it asks before posting anything.


.claude/skills/gini-review/references/comment-style.md
```diff
@@ -0,0 +1,204 @@
+# Comment format
```
ValentinaIancu-Gini, line 8: [Suggestion] @MozhganPeivandianSharbaf we normally have, at least on iOS, a style for the comments like:

[Suggestion]<message here>,
[Request]<message here>,
[Clarification] <message here>.

Can we introduce it here?

.claude/skills/gini-review/references/platform-rules.md
```diff
@@ -0,0 +1,102 @@
+# Writing a platform layer
```
ValentinaIancu-Gini, line 8: [Clarification] @MozhganPeivandianSharbaf I was thinking that in this part we can use the agent team to analyse the PR from a technical perspective. Can we do this?

.claude/skills/gini-review/pr-review/references/ticket-context.md
```diff
@@ -0,0 +1,177 @@
...
+The ticket is the specification. Reviewing a diff without it verifies only that the code is
+*well-formed*, not that it is *correct*. This file covers how to fetch it, what to extract, and how
+to turn it into logic checks.
+
+## Fetching
+
+Site is **`ginis.atlassian.net`** (note the trailing `s` — `gini.atlassian.net` returns
+403 "app is not installed on this instance"). Use the Atlassian MCP tool `getJiraIssue`:
+
+- `cloudId`: `7740065a-6c74-4abe-89b4-eed057e702d4`
```
ValentinaIancu-Gini: [Clarification] @MozhganPeivandianSharbaf What CloudId is this? did we leak something here?

.claude/skills/gini-review/pr-review/references/ticket-context.md
```diff
@@ -0,0 +1,177 @@
...
+- **Logic checks** — the 8-step method, including the opposite-direction check
+
+**Does not cover:** repo coding rules and published API surface → the repo's platform layer
+(`../SKILL.md` §0) · comment wording → `comment-style.md`
+
+The ticket is the specification. Reviewing a diff without it verifies only that the code is
+*well-formed*, not that it is *correct*. This file covers how to fetch it, what to extract, and how
+to turn it into logic checks.
+
+## Fetching
```
ValentinaIancu-Gini: [Suggestion] I think we can improve this and not necessarily rely on Jira but also on the user input, like copy-paste from the ticket in case the connection doesn't work or someone doesn't have the Jira MCP.

.claude/skills/gini-review/references/comment-style.md
MozhganPeivandianSharbaf: Done — added a "three categories" section using exactly those labels, and every posted comment now opens with one. The report sections are renamed to Requests / Suggestions / Clarifications so they map straight onto them.

.claude/skills/gini-review/references/platform-rules.md
MozhganPeivandianSharbaf: Yes — added as an opt-in pass in `SKILL.md` §3. On a big diff (>15 files, >500 lines, or more than one module) it asks first, then fans out one subagent per dimension over the whole diff, and a second pass tries to refute each finding before it reaches the report.

.claude/skills/gini-review/pr-review/references/ticket-context.md
MozhganPeivandianSharbaf: No leak — it is the workspace id for `ginis.atlassian.net`, not a credential, and it does nothing without an authorised OAuth session. Removed it anyway: the skill now resolves it at run time via `getAccessibleAtlassianResources`.

.claude/skills/gini-review/pr-review/references/ticket-context.md
MozhganPeivandianSharbaf: Added — new section "When Jira is not available". If the fetch fails it asks you to paste the ticket and reviews against that, marking it "pasted, not fetched" in the report. A bare URL is rejected (Jira URLs return the login page) and "skip" stays an explicit option.

.claude/skills/gini-review/references/comment-style.md
```diff
@@ -0,0 +1,232 @@
...
+2. **State the consequence.** What goes wrong for whom — the user, the consumer, the next maintainer.
+   This is the clause that justifies the comment existing.
+3. **Give a concrete fix.** Name the API, file, or pattern. For a `[Clarification]`, this is the
+   question instead — one question, answerable in a sentence.
+
+Three comments in this shape, as a calibration target:
+
+> `[Request]` The snippet under "Self-managed authentication" uses Kotlin syntax (`val`, lambda), but the
+> code block is marked as `java`, which leads to misleading syntax highlighting and copy/paste for
+> readers. Either switch the directive to `kotlin` or rewrite the snippet in Java syntax.
```
zladzeyka, line 125: [Clarification]: Does it make sense to mention `Swift` since it's inside comment style or move platform specific here is a programming language inside `platform.md`

.claude/skills/gini-review/references/comment-style.md
```diff
@@ -0,0 +1,232 @@
...
+  so rather than raising it.
+- **Cap the list.** A handful at most. A long optional list reads as noise and buries the requests.
+
+## Posting to GitHub
+
+For mechanical changes, attach a suggestion block so the author gets one-click apply:
+
+````markdown
+[Request] The code block is marked `java` but the snippet is Kotlin, which gives readers wrong
+highlighting. Switch the directive to `kotlin`.
```
zladzeyka, line 218: [Clarification]: Does it make sense to mention Swift since it's inside comment style or move platform specific here is a programming language inside platform.md

.claude/skills/gini-review/references/ticket-context.md
```diff
@@ -0,0 +1,216 @@
...
+403 "app is not installed on this instance"). Use the Atlassian MCP tool `getJiraIssue`:
+
+- `cloudId`: **resolve it at run time — do not hard-code it.** Call
+  `getAccessibleAtlassianResources` and take the id of the `ginis.atlassian.net` entry. It is a
+  workspace identifier, not a credential: it is inert without an authorised OAuth session, and every
+  Atlassian client discovers it the same way. Looking it up also keeps this file correct if the site is
+  ever migrated.
+- `issueIdOrKey`: the key from the branch name — the **first match of `[A-Z]{2,5}-[0-9]+` anywhere in
+  the name**, not anchored at the start. Branches are usually `<TICKET>-<kebab-description>`, but the
+  key can sit under a segment (`backup/PP-1234-…`, `feature/FEAT-001-…`), and an anchored pattern
```
zladzeyka, line 51: [Clarification]: I've seen we used `feature/PP-1234 1-…`also `release/P-12341-…` but backup is new shall we remove it?

.claude/skills/gini-review/platform.md
```diff
@@ -0,0 +1,468 @@
...
+file is read on every review in this repository, so it holds only standing conventions. Ticket keys,
+version numbers and API names that appear below are **format placeholders**, never live values — do not
+carry one into a review as if it came from the PR. Where a count or a version is quoted, it is
+approximate and as-of-writing: re-check it against the tree rather than citing it as fact.
+
+**Where those instructions have drifted from the code.** Cite them anyway — they are what the team agreed
+— but know the gaps, so a finding doesn't collapse under one reply from the author:
+
+- `CLAUDE.md` mandates the `UIColor.GiniBank.*` / `UIColor.GiniCapture.*` colour namespaces, but a newer
+  `GiniColorScheme` family has reached ~25 files without the doc being updated. See §5.
```
zladzeyka, line 40: [Suggestion]:The GiniColorScheme system, defined in GiniUtilites and applied via giniBankColorScheme() / giniHealthColorScheme() and .colorScheme.* accessors

.claude/skills/gini-review/platform.md
```diff
@@ -0,0 +1,468 @@
...
+leaks by accident here. That flips what review is for: every `public` or `open` in the diff is a
+deliberate decision, so review it as one rather than hunting for accidents.
+
+- Visibility ladder: `open` > `public` > `package` > `internal` (default) > `fileprivate` > `private`.
+- **All seven modules ship to a public release repo** — there is no internal-only module:
+  `GiniBankSDK` → `gini/bank-sdk-ios`, `GiniCaptureSDK` → `gini/capture-sdk-ios`, `GiniHealthSDK` →
+  `gini/health-sdk-ios`, `GiniBankAPILibrary` → `gini/bank-api-library-ios`, `GiniHealthAPILibrary` →
+  `gini/health-api-library-ios`, `GiniUtilites` → `gini/utilites-ios`, and — **despite the name** —
+  `GiniInternalPaymentSDK` → `gini/internal-payment-sdk-ios`. Treat a new `public` symbol in
+  `GiniInternalPaymentSDK` as published API, not as internal shared code.
```
zladzeyka, line 64: [Request]: `GiniInternalPaymentSDK` and `GiniUtilites` are internal SDKs and their public actually should be considered for now as internal shared code

.claude/skills/gini-review/platform.md
```diff
@@ -0,0 +1,468 @@
...
+
+So a PR touching **only `GiniUtilites`** runs `utilites.check.yml` and nothing else — no dependent
+SDK's tests run, even though every SDK depends on it. Same for `GiniInternalPaymentSDK`: HealthSDK's
+suite does not run. On such a PR, **the reviewer is the only gate on downstream breakage.** Trace the
+changed symbol's call sites into each dependent and say in the report that you did — or that CI did
+not.
+
+## 5. Architecture and style
+
+`CLAUDE.md` §"MyApp Standards" states these as **MUST**s, which makes them the strongest citations
```
zladzeyka, line 147: [Suggestion]: We already have merged architecture and style I suggest to refer here to the existing sources

.claude/skills/gini-review/platform.md
```diff
@@ -0,0 +1,468 @@
...
+
+```
+<type>(<project>): <subject>
+
+<body>
+
+<ticket-id>
+```
+
+- Types: `feat`, `fix`, `refactor`, `ci`, and **`ai`** — the last one covers Claude Code skills, agents
```
zladzeyka, line 239: [Request]: Please refer here to the existing Commit style

.claude/skills/gini-review/platform.md
```diff
@@ -0,0 +1,468 @@
...
+
+Practical consequence: **source compatibility is what you review.** Library evolution buys resilience
+for adding stored properties to non-frozen public structs, but it does not make a renamed method or a
+changed signature safe for anybody.
+
+### Source-breaking in Swift — the list worth checking
+
+Every one of these compiles fine here and breaks an integrator:
+
+- **Removing or renaming** any `public`/`open` symbol, or changing its access level downward.
```
zladzeyka, line 385: [Clarification]: We have  3 open class declarations (GiniBankNetworkingScreenApiCoordinator, GiniScreenAPICoordinator, BottomSheetViewController) plus 2 open override func in BottomSheetViewController
  — at least 5 real open declarations.

Shall we mention here considering the new code only?

zladzeyka (dismissed):
Thank you for PR

.claude/skills/gini-review/references/comment-style.md
```diff
@@ -0,0 +1,232 @@
...
+
```
ValentinaIancu-Gini: [Request] Please remove this. Test

ValentinaIancu-Gini (dismissed):

abolfazlimahdi (approved):

.claude/skills/gini-review/references/comment-style.md
MozhganPeivandianSharbaf: Agreed. The examples are language-neutral now, and I added a note at the top of that section: anything naming a language, linter or framework belongs in the repo's `platform.md`, not in a shared file.

.claude/skills/gini-review/references/comment-style.md
MozhganPeivandianSharbaf: Same fix here — the suggestion-block example no longer uses Swift or Kotlin either.

.claude/skills/gini-review/references/ticket-context.md
MozhganPeivandianSharbaf: Removed. I checked the head branches of the last 100 PRs in both repos: `fix/` and `feat/` are iOS-only (Android has zero), and only `release/`, `copilot/` and `dependabot/` are common to both. So I dropped the prefix list entirely — the file now just says to match on start-or-slash and let any segment through, which holds on both platforms and can't go stale.

.claude/skills/gini-review/platform.md
MozhganPeivandianSharbaf: Updated, with one correction: there is no `giniHealthColorScheme()` and no `.colorScheme.*` accessor in the tree — `giniBankColorScheme()` in `GiniBankColorScheme.swift` is the only factory today, so the text says that. It also points at `.claude/rules/mandatory-rules.md` for the preference order, and I replaced the file counts with the grep that produces them, since one had already drifted.

.claude/skills/gini-review/platform.md
MozhganPeivandianSharbaf: Done. Split into the five customer-facing modules and these two as internal packages — their `public` surface is treated as internal shared code, so needing a new `public` symbol there is no longer a finding on its own. It only becomes one if the symbol is re-exported through a customer-facing SDK. §9 carries the same distinction.

.claude/skills/gini-review/platform.md
MozhganPeivandianSharbaf: Done — §5 now points at `.claude/rules/mandatory-rules.md` as the merged source, with `CLAUDE.md`/`AGENTS.md` winning on drift. The list underneath is just a reviewer's index of which rules get broken most often, not a second source.

.claude/skills/gini-review/platform.md
MozhganPeivandianSharbaf: Done — §7 now refers to `.github/instructions/commit-message-guideline.instructions.md` for the format and to `.git-stuff/commit-msg-template.txt` for the allowed `<type>` values. I dropped the inline list of types, since that is what went stale.

.claude/skills/gini-review/platform.md
MozhganPeivandianSharbaf: Fixed to five, named individually, and noted that grepping `open` also hits the `open url:` argument label in `GiniBankUtils.swift`, which is not a declaration. And yes — the new-code-only scope is now explicit for `open`: a pre-existing one is not a finding.

zladzeyka (approved):
Thank you @MozhganPeivandianSharbaf for addressing the comments, looks good to me.

PR #1232, opened 2026-08-05
ai: Add gini-release skill for iOS package releases

MozhganPeivandianSharbaf:
## Pull Request Description

Adds `/gini-release`, a Claude Code skill that guides an iOS package release end-to-end: it asks for the packages and their new versions, creates the Jira RC ticket(s) in PP/HEAL, bumps every version site in dependency order with one commit per package, gates the tag push behind QA sign-off, and walks through the post-tag checklist. It exists so a release follows one documented path instead of being re-derived each time from a 15-line `RELEASE.md` that is partly out of date.

N/A 

The skill has the same mission as the Android repo's `gini-release`, but every step is written against what this repo actually does. The parts with no Android equivalent are where the risk sits: a release here touches **up to four** places per package — the `<Package>Version.swift` constant, every dependent's `Package-release.swift` `.exact()` pin, `Documentation/source/Installation.md` where the package has one, and the version-carrying Figma URLs in the Health docs (routed through the existing `/update-figma-links` skill rather than hand-edited). Tag creation stays behind an explicit confirmation, since a pushed tag immediately force-publishes the package into its public release repo.

## Notes for Reviewers

- **Skill-only change** — one new file under `.claude/skills/gini-release/`. No SDK, example app, workflow or CI file is touched, so no tests apply and no module is affected.
- **Derived from**, and worth checking against: `RELEASE.md`, `RELEASE-ORDER.md`, the fastlane lanes (`create_release_tags`, `create_documentation_release_tags`, `publish_podspec`, `util/versions.rb`, `util/swift_package_releases.rb`), the `*.release.yml` workflows, the release-tag and bump-commit history, and the last ~23 `Release Candidate` tickets in PP and HEAL.
- **Three load-bearing claims** that a release owner should confirm read correctly:
  - **The bank chain is version-locked.** `GiniBankSDK` and `GiniCaptureSDK` tagged the same version on every 4.x release (4.0.0, 4.1.0, 4.1.1, 4.2.0, 4.2.2, 4.3.0); `GiniBankAPILibrary` joins on minor/major but sits out patches (no 4.1.1 or 4.2.2 tag). The **health chain is not locked** — `GiniHealthSDK` released 5.6.1 and 6.1.0 alone. The skill flags a list that violates this instead of silently "fixing" it.
  - **The `Package-release.swift` pins are the silent failure.** The PR checks only resolve the local `Package.swift`, so a missed `.exact()` pin breaks the release build with nothing catching it beforehand.
  - **All seven release repos publish GitHub releases**, including `utilites-ios` and `internal-payment-sdk-ios` — whereas Jira fix versions cover only the customer-facing products (the 4.4.0 bank release carried three, none for `GiniUtilites`).
- **Four places `RELEASE.md` is stale** are documented in a closing section rather than worked around. Worth deciding whether to fix `RELEASE.md` itself in a follow-up so the two don't drift further:
  - *"don't forget to update also extended package with Pinning!"* — **there are no Pinning packages.** No `Package.swift` declares one and no `*PinningVersion.swift` exists; SSL pinning is now a source folder inside the main packages. The `ignored_packages` rule in `create_release_tags`, the `GiniBankSDKPinning` comment in `bank-sdk.check.yml` and the `capture-sdk-pinning-ios` reference are all vestigial.
  - *"Make sure you are on the `main` branch"* — releases in the history actually run from a `release/<theme>` branch (`release/qr-code-improvements`, `release/liquid_glass_bank_sdk`, `release/bank_sdk_release_4.2.2`) and reach `main` by PR.
  - *"documentation file"* is unnamed — it is `Documentation/source/Installation.md`, and only `GiniBankAPILibrary` and `GiniHealthAPILibrary` have one.
  - `RELEASE-ORDER.md` **is hand-maintained here.** No lane generates it, which is the opposite of Android, where the equivalent file is generated and must never be edited by hand.
- **Not yet exercised on a real release.** Steps 1 and 4 are verified against the repo and the tag history; step 2's Jira mechanics and steps 5–7 are written from the ticket conventions and the Android equivalent. The next actual release is the real test — please correct the file in place rather than working around it.
- **One deliberate choice:** step 1 asks for the full version list rather than deriving Capture/Bank SDK from the version lock, because `GiniBankAPILibrary`'s patch behaviour is a judgement call and guessing it wrong silently is worse than one extra question. Easy to change if you'd rather it proposed the set for confirmation.


.claude/skills/gini-release/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+
+So **ask which branch this release ships from**; do not assume. Two cases:
+
+- **An existing `release/<theme>` branch** — the normal case when the release has a theme. Confirm it is
+  up to date with `main` before bumping.
+- **`main` directly** — for a patch release with no integration branch.
+
+Do not invent an Android-style `PP-XXX-RC-…` branch; that convention does not exist in this repo.
+
+## 4. Bump versions, one commit per package, in release order
```
ValentinaIancu-Gini: [Suggestion] @MozhganPeivandianSharbaf we normally don't bump the version until we receive the go from our QA. So I would exclude this step. I would prefer to keep it last in the process since, normally, on that commit, we add tags for release and is more meaningful for us to match the version bump with the tag.

.claude/skills/gini-release/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+package into its public release repo.
+
+## 1. Ask for the packages and their new versions
+
+Ask the user for the packages being released with their **new versions**, one per line:
+
+```
+GiniCaptureSDK 4.4.1
+GiniBankSDK 4.4.1
+GiniUtilites 2.6.0
```
ValentinaIancu-Gini, line 24: [Suggestion] For this, we need to add the `GiniBankAPILibrary` also.

.claude/skills/gini-release/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+| `GiniHealthAPILibrary` | `HealthAPILibrary/GiniHealthAPILibrary/Sources/GiniHealthAPILibrary/GiniHealthAPILibraryVersion.swift` | `gini/health-api-library-ios` |
+| `GiniUtilites` | `GiniComponents/Utilities/GiniUtilites/Sources/GiniUtilites/GiniUtilitesVersion.swift` | `gini/utilites-ios` |
+| `GiniCaptureSDK` | `CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/GiniCaptureSDKVersion.swift` | `gini/capture-sdk-ios` |
+| `GiniInternalPaymentSDK` | `GiniComponents/InternalPaymentSDK/GiniInternalPaymentSDK/Sources/GiniInternalPaymentSDK/GiniInternalPaymentSDKVersion.swift` | `gini/internal-payment-sdk-ios` |
+| `GiniBankSDK` | `BankSDK/GiniBankSDK/Sources/GiniBankSDK/GiniBankSDKVersion.swift` | `gini/bank-sdk-ios` |
+| `GiniHealthSDK` | `HealthSDK/GiniHealthSDK/Sources/GiniHealthSDK/GiniHealthSDKVersion.swift` | `gini/health-sdk-ios` |
+
+**Spell `GiniUtilites` with one `i` in the middle** — that is the real package name. Past commits and one
+tag used `GiniUtilities`, and a mis-spelled tag is invisible to `create_release_tags`, which derives the
```
ValentinaIancu-Gini: [Suggestion] We know we have a mis spelling issue but still I want the tag to mirror the SDK name so I would remove the second sentence here or at least to modify to say is ok to have the tag  as the SDK name

.claude/skills/gini-release/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+
+- **`GiniBankSDK` and `GiniCaptureSDK` share a version number.** Every bank release in the 4.x line —
+  4.0.0, 4.1.0, 4.1.1, 4.2.0, 4.2.2, 4.3.0 — tagged both at the same version. If the user lists one
+  without the other, or at a different number, flag it. `GiniBankAPILibrary` joins the lock on **minor and
+  major** releases (4.0.0, 4.1.0, 4.2.0, 4.3.0) but usually sits out **patches** (it has no 4.1.1 or
+  4.2.2 tag).
+- **The health chain is not locked.** `GiniHealthSDK` frequently releases alone (5.6.1, 6.1.0);
+  `GiniHealthAPILibrary` joined only at 6.0.0. Do not "correct" a health list to match.
+- **`GiniUtilites` and `GiniInternalPaymentSDK` have their own version lines** (currently 2.x and 3.x) and
+  never share the SDK numbers. A list mixing them at an SDK's version is a mistake.
```
ValentinaIancu-Gini: [Clarification] I'm not sure if this info is relevant.

.claude/skills/gini-release/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+`customfield_*` named `Sprint`, and set the entry whose `state` is `active`. The active sprint may belong
+to a board other than the project's own — if so the ticket legitimately won't show on the project board;
+point at the `boardId` from the sprint record. JQL against a non-existent value returns an empty result
+rather than an error, so "no results" never proves absence.
+
+**Drive the ticket's status yourself.** Move it to `In Progress` when the bumps start and to the
+waiting-for-QA status at the step 5 gate, via `transitionJiraIssue`. **Resolve transitions per ticket by
+name, case-insensitively** — call `getTransitionsForJiraIssue` on the actual ticket; the same status is
+spelled differently between the two projects, and ids collide across projects, so a copied id can silently
+move a ticket to `Cancelled`.
```
ValentinaIancu-Gini: [Clarification] Maybe this should be changed after the bump section is removed. I would add instead a slack notification for when the ticket is created 😅 if is possible.

.claude/skills/gini-release/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+`RELEASE.md` says to be on `main`. **In practice most releases happen on a long-lived
+`release/<theme>` branch** that feature PRs have been merged into — e.g. `release/qr-code-improvements`,
+`release/liquid_glass_bank_sdk`, `release/bank_sdk_release_4.2.2`. Naming is not standardised, and the
+theme usually matches the suffix on the Jira fix versions.
+
+So **ask which branch this release ships from**; do not assume. Two cases:
+
+- **An existing `release/<theme>` branch** — the normal case when the release has a theme. Confirm it is
+  up to date with `main` before bumping.
+- **`main` directly** — for a patch release with no integration branch.
```
ValentinaIancu-Gini: [Suggestion] I would remove this. I would not create an RC directly from main. Main should be untouched and we should always have a release branch.

.claude/skills/gini-release/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+
+Do not invent an Android-style `PP-XXX-RC-…` branch; that convention does not exist in this repo.
```
ValentinaIancu-Gini: [Clarification] Is this needed?

.claude/skills/gini-release/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+
+If the release ships from a `release/<theme>` branch, open the PR into `main` now — the bumps merge like
+any other change, and tags are created after that lands.
+
+## 5. Wait for QA — hard gate
+
+Stop here. Tags may only be created after QA signs the RC ticket off and assigns it back. Ask the user to
+confirm; never infer it. The Firebase build QA tests is the one linked in the RC ticket.
+
+## 6. Create and push release tags
```
ValentinaIancu-Gini: [Clarification] Why do we have more sections after 5? How are these triggered if we stop for QA? Also, I would control the create and push tags a bit and still trigger it manually in the terminal, since it's just one command we need to run.
iOS steps after pushing tags are complicated at the moment, and I would not suggest having an automated process since we always have problems running some commands.

ValentinaIancu-Gini:
@zladzeyka let's discuss this. I think we need to take it over since it's our release process.

.claude/skills/gini-release/SKILL.md
zladzeyka: [Suggestion]: also add GiniHealthAPILibrary, GiniInternalPaymentSDK and GiniHealthSDK?

.claude/skills/gini-release/SKILL.md
zladzeyka: [Suggestion]: follow misspelled name GiniUtilites;2.4.0 and it can be fixed in the next major

.claude/skills/gini-release/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+tag used `GiniUtilities`, and a mis-spelled tag is invisible to `create_release_tags`, which derives the
+name from the `*Version.swift` filename.
+
+Sanity-check the list against the current versions in those files and flag — don't silently fix:
+
+- **`GiniBankSDK` and `GiniCaptureSDK` share a version number.** Every bank release in the 4.x line —
+  4.0.0, 4.1.0, 4.1.1, 4.2.0, 4.2.2, 4.3.0 — tagged both at the same version. If the user lists one
+  without the other, or at a different number, flag it. `GiniBankAPILibrary` joins the lock on **minor and
+  major** releases (4.0.0, 4.1.0, 4.2.0, 4.3.0) but usually sits out **patches** (it has no 4.1.1 or
+  4.2.2 tag).
```
zladzeyka, line 58: [Request]: Please reformulate it. We don't have it as an agreement versions are defined based on the changes in the source code and can be different for *`GiniBankSDK` and `GiniCaptureSDK`

.claude/skills/gini-release/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+<RC-ticket-id>
+```
+
+`<Package>` is the package name (`GiniBankSDK`). Use the ticket of that package's side; for
+`GiniUtilites` in a both-sides release include both ticket ids. The valid `type` values live in
+`.git-stuff/commit-msg-template.txt` — note the history contains `feature(...)`, which is **not** valid;
+use `feat`.
+
+Then run `make lint scheme=<Scheme>` for each affected package (the `AGENTS.md` gate) plus the touched
+packages' unit tests, and push the branch. **No tags yet.**
```
zladzeyka, line 225: [Suggestion]: Introduce fix linting ->   swiftlint --fix BankSDK/GiniBankSDK/Sources

.claude/skills/gini-release/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+
+If the release ships from a `release/<theme>` branch, open the PR into `main` now — the bumps merge like
```
zladzeyka, line 227: [Clarification]: we don't bump on this stage do we?

.claude/skills/gini-release/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+
+- Tag format is strict: `<Package>;X.Y.Z`, or `<Package>;X.Y.Z-betaNN` with **exactly two** beta digits.
+  Anything else is ignored by the release workflows.
+- **Each pushed tag immediately triggers that package's release workflow**, which runs the package's
+  checks, then clones the release repo, wipes it, copies the package in with `Package-release.swift`
+  renamed to `Package.swift`, commits and tags it. Only push when the release is truly go.
+- Verify the workflows started under GitHub Actions afterwards. The release workflow also publishes that
+  package's Jazzy documentation as a dependent job.
+
+**XCFramework builds are separate and not part of this.** The `build-xcframeworks` job is commented out in
```
zladzeyka, line 254: [Clarification]: Why we don't have a platform specific file for this skill?

.claude/skills/gini-release/SKILL.md
```diff
@@ -0,0 +1,298 @@
...
+   including `utilites-ios` and `internal-payment-sdk-ios`, so do not skip a package for being
+   "internal" — unlike Jira fix versions, which only cover the customer-facing products. Use the markdown
+   release notes from the Jira release description.
+2. **Jira** — confirm each release has its tickets connected via "Fix versions" and notes in the
+   description, then publish the releases in PP/HEAL.
+3. **Docs** — the release workflow publishes Jazzy automatically. A **documentation-only** change between
+   versions is released separately with `bundle exec fastlane create_documentation_release_tags`, which
+   tags `<Package>;<version>;doc-<n>`.
+4. **CocoaPods** — only `GiniBankSDK` has a podspec, and **no workflow publishes it**. It is a manual
+   `bundle exec fastlane publish_podspec` that requires the XCFrameworks from step 6's note and pushes to
```
zladzeyka, line 271: [Request]: Clarification is needed here how to run this command 
`bundle exec fastlane publish_podspec xcframeworks_folder_path:</path/to/xcframeworks> pod_name:<Name> podspecs_repo_sdk_folder_path:</path/to/gini-podspecs> template_podspec_path:</path/to/Pod/<Pod_Name.podspec> replacing with the correct paths and name. ` follow https://ginis.atlassian.net/wiki/spaces/PLMO/pages/83689511/Mobile+Release+Process?xpis=eyJicmlkZ2UiOiJxdWlja0ZpbmQiLCJpZCI6IjE3ODcxNTUyNTY2MTMiLCJzb3VyY2UiOiJjb25mbHVlbmNlIn0%3D  17 section for the details

zladzeyka (changes requested):
Thank you for the PR.
I would go with steps and use corresponded source and points from https://ginis.atlassian.net/wiki/spaces/PLMO/pages/83689511/Mobile+Release+Process?xpis=eyJicmlkZ2UiOiJxdWlja0ZpbmQiLCJpZCI6IjE3ODcxNTUyNTY2MTMiLCJzb3VyY2UiOiJjb25mbHVlbmNlIn0%3D

.claude/skills/gini-release/SKILL.md
ValentinaIancu-Gini: the make lint here is not same thing as swiftlint. It should just compile the project and make sure we don't have errors during compilation.

zladzeyka (approved):

PR #1244, opened 2026-08-19
Release Q1 2026: Credit note handling and native bottom sheets (GiniBankSDK 4.5)

ValentinaIancu-Gini:
## Pull Request Description

This PR brings the Q1 2026 Gini Bank release work into `release/GiniBankSDK-4.5`. It delivers the credit-note handling feature and the native bottom-sheet rework, plus supporting test hardening.

[PP-2263]

**What's included:**

- **Credit note support (PP-2263):** `ClientConfiguration` gains `creditNoteHintEnabled`; when a document is marked as a credit note and the hints are enabled, the SDK shows a new `CreditNoteWarningViewController` bottom sheet (Cancel / Proceed) and strips the credit-note compound extractions from the transfer summary on delivery (`excludingCompoundExtractions`, chained with `excludingAmountToPay`). Spec in `specs/PP-2263-feature.md`.
- **Native bottom-sheet rework (PP-2232):** the hand-rolled `GiniBottomSheetViewController` is replaced by `UISheetPresentationController` via the `GiniBottomSheetPresentable` protocol, now shared from `GiniUtilites` together with the new `GiniBottomSheetAccessibilityRestorable` protocol for VoiceOver focus restoration after dismissal. The Edit Line Item screen was refactored onto the native sheet.
- **Layout DSL relocation:** the Gini constraint DSL (`giniMakeConstraints` and friends) moved from `GiniCaptureSDK/Core/Layout` to `GiniUtilites/Layout` so all SDKs can use it.
- **Test hardening:** shared-configuration test suites serialized (`GiniConfigurationSharedStateSuite`), transfer-summary integration tests fixed against cross-test contamination (polling + per-test document-service re-pinning), new credit-note fixtures and decoding tests.
- **Example app:** credentials are resolved per selected credential set and API environment; a settings toggle for the credit-note hint was added.

## Notes for Reviewers

- Integration tests require `TEST_CLIENT_ID` / `TEST_CLIENT_SECRET` environment variables (transfer-summary + client-configuration suites hit the real API).
- Unit tests added/updated: `ClientConfigurationTests`, `ExtractionsContainerTest` (credit-note fixture), `NetworkingScreenApiCoordinatorTests` (gating + stripping helpers), `GiniBankConfigurationTests`, `RemoteConfigPropagationTests`, `SettingsViewControllerTests`.
- Suggested manual check: analyze a document marked as a credit note with the hint flags enabled — the warning sheet should appear; "Proceed" delivers the transfer summary without the credit-note compound extractions, "Cancel" returns to the flow. Verify VoiceOver focus is trapped in the sheet while presented.
- Known follow-ups from the internal multi-agent review (tracked separately, not blocking this PR): decode tolerance for absent `creditNoteHintEnabled` in cached/older backend responses, re-export of the moved public layout DSL symbols from `GiniCaptureSDK`, and VoiceOver reachability of the Edit Line Item validation errors.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[PP-2263]: https://ginis.atlassian.net/browse/PP-2263?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
```diff
@@ -373,7 +373,7 @@ extension GiniBankNetworkingScreenApiCoordinator {
             if !document.isReviewable {
                 uploadAndStartAnalysisWithReturnAssistant(document: document,
                                                           networkDelegate: networkDelegate,
-                                                          uploadDidFail: {
+                                                          uploadDidFail:{
```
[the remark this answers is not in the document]
ValentinaIancu-Gini: This is ok.

GiniComponents/Utilities/GiniUtilites/Sources/GiniUtilites/Protocols/BottomSheet/GiniBottomSheetAccessibilityRestorable.swift
```diff
@@ -0,0 +1,69 @@
...
+    var presenterViewController: UIViewController { get }
+
+    /// The view controller to focus on after bottom sheet dismissal
+    var accessibilityFocusTargetViewController: UIViewController? { get }
+}
+
+extension GiniBottomSheetAccessibilityRestorable {
+
+    /// Default implementation: focus on the presenter's visible content
+    public var accessibilityFocusTargetViewController: UIViewController? {
```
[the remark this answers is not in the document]
ValentinaIancu-Gini: will do

BankSDK/GiniBankSDKExample/Tests/UnitTests/SettingsViewControllerTests.swift
```diff
@@ -1588,6 +1588,46 @@ extension SettingsViewModelTests {
...
+        if case .switchOption(var data) = contentData[index.section].items[index.row] {
+            guard data.type == .creditNoteHintEnabled else {
+                XCTFail("Expected type `creditNoteHintEnabled`, found a different one: \(data.type)")
+                return
+            }
+            data.isSwitchOn = true
+            configuration.creditNoteHintEnabled = data.isSwitchOn
+
+            XCTAssertTrue(configuration.creditNoteHintEnabled,
+                          "credit note feature should not be enabled in the gini configuration")
```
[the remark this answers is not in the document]
ValentinaIancu-Gini: will do

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/ReturnAssistant/EditLineItem/EditLineItemView.swift
```diff
@@ -218,59 +192,67 @@ final class EditLineItemView: UIView {
...
+            $0.top.equalToSuperview().constant(Constants.verticalPadding)
+            $0.height.greaterThanOrEqualTo(Constants.headerButtonMinimumHeight)
+        }
+
+        // Title label
+        titleLabel.giniMakeConstraints {
+            $0.centerY.equalTo(cancelButton)
+            $0.centerX.equalToSuperview()
+            $0.leading.greaterThanOrEqualTo(cancelButton.trailing).constant(Constants.titlePadding)
+            $0.trailing.lessThanOrEqualTo(saveButton.leading).constant(Constants.titlePadding)
```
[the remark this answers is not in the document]
ValentinaIancu-Gini: it is ok as it is.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Custom views/ButtonsView.swift
```diff
@@ -64,14 +78,23 @@ class ButtonsView: UIView {
...
+    private func configureStackViewButton(title: String) -> MultilineTitleButton {
+        let button = MultilineTitleButton()
+        button.setTitle(title, for: .normal)
+        button.titleLabel?.font = giniConfiguration.textStyleFonts[.bodyBold]
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.accessibilityLabel = title
+
+        NSLayoutConstraint.activate([
+            button.heightAnchor.constraint(greaterThanOrEqualToConstant: Constants.buttonMinimumHeight)
+        ])
```
zladzeyka:
```suggestion
        button.giniMakeConstraints {
          $0.height.greaterThanOrEqualTo(Constants.buttonMinimumHeight)
      }
```

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Custom views/ButtonsView.swift
```diff
@@ -38,13 +27,27 @@ class ButtonsView: UIView {
...
+    }
+
+    private func setupView() {
         addSubview(buttonsView)
         configureButtons()
+        arrangeButtons()
+
         NSLayoutConstraint.activate([
             buttonsView.leadingAnchor.constraint(equalTo: leadingAnchor),
             buttonsView.trailingAnchor.constraint(equalTo: trailingAnchor),
```
zladzeyka:
```suggestion
buttonsView.giniMakeConstraints {
     $0.edges.equalToSuperview()
}
```

zladzeyka (commented):
## Accessibility follow-up: VoiceOver focus leak on the two warning sheets

### What's happening today

`presentAsBottomSheet(from:)` (in `GiniUtilites/.../GiniBottomSheetPresentable.swift:162-179`) is a **one-way accessibility trap**. On presentation it does three things — set `presenter.view.accessibilityElementsHidden = true`, set `self.view.accessibilityViewIsModal = true`, and post `.screenChanged` — to keep VoiceOver focus locked inside the sheet. That is correct. But the helper does not observe dismissal at all. Nothing writes `accessibilityElementsHidden = false` back. The trap is only ever lifted by the caller, via `GiniBottomSheetAccessibilityRestorable.restoreAccessibilityAfterBottomSheetDismissal()`.

### What breaks

`presentDocumentMarkedAsPaidBottomSheet` and `presentDocumentMarkedAsCreditNoteBottomSheet` (`GiniBankNetworkingScreenApiCoordinator.swift:1058-1092`) never wire that restore hook, and `InfoBottomSheetViewController` doesn't even expose an `onDismiss` property to wire. So after a VoiceOver user taps **Cancel** or **Proceed**:

- `screenAPINavigationController.view.accessibilityElementsHidden` stays `true` for the remainder of the coordinator's lifetime.
- VoiceOver swipes on the presenter hit nothing — the view reports zero accessible elements, so the user hears silence or falls through to background elements.
- Any screen subsequently presented on the same nav controller (e.g. the transaction-docs alert reached on the credit-note Proceed path via `presentTransactionDocsAlert` → `handleTransactionDocsAlert`) inherits an invisible-to-VoiceOver container. The nav bar is also affected — `restoreAccessibilityAfterBottomSheetDismissal` explicitly resets `navigationBar.accessibilityElementsHidden`, confirming the flag propagates to it.

### Why this only shows on VoiceOver

Sighted QA sees the sheet slide down cleanly and moves on — there is no visual regression. The bug is entirely in the accessibility tree, which is why it slipped through pre-merge testing.

### Why `DigitalInvoiceCoordinator` works

It already implements both halves for the Edit Line Item sheet: (a) `EditLineItemViewController` declares `var onDismiss: (() -> Void)?` and fires it in `viewWillDisappear` (`EditLineItemViewController.swift:39, 71-76`); (b) the coordinator conforms to `GiniBottomSheetAccessibilityRestorable` and assigns `viewController.onDismiss = { restoreAccessibilityAfterBottomSheetDismissal() }` (`DigitalInvoiceCoordinator.swift:116-119, 154-157`). The Bank coordinator's two warning-sheet paths skipped both halves.

### The fix — six edits

Two are inlined below as one-click `suggestion` blocks (the diff hunks reach them). The other four are outside any diff hunk on this PR, so GitHub can't render them as inline suggestions — they're listed here with `[Suggestion]:` + file:line + copy-pasteable snippets.

---

**[Suggestion]: Expose `onDismiss` on `InfoBottomSheetViewController`.**
`CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Custom views/InfoBottomSheet/InfoBottomSheetViewController.swift`, after line 101 (end of the `GiniBottomSheetPresentable protocol` MARK section):

```swift
    /// Called when the bottom sheet is dismissed (by any means).
    public var onDismiss: (() -> Void)?
```

Same shape as `EditLineItemViewController.swift:39`. Both `CreditNoteWarningViewController` and `DocumentMarkedAsPaidViewController` subclass this VC and inherit the hook for free.

---

**[Suggestion]: Fire `onDismiss` from `viewWillDisappear`.**
Same file, after the `viewDidLoad` override (line 121):

```swift
    public override func viewWillDisappear(_ animated: Bool) {
        super.viewWillDisappear(animated)
        onDismiss?()
    }
```

This is the mechanical trigger the coordinator uses to call `restoreAccessibilityAfterBottomSheetDismissal()` right after the sheet slides away.

---

**[Suggestion]: Add `import GiniUtilites` in the coordinator.**
`BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift`, alongside the existing imports (around line 9):

```swift
import GiniUtilites
```

`GiniBottomSheetAccessibilityRestorable` was moved into `GiniUtilites` in this PR; the conformance below needs the explicit import.

---

**[Suggestion]: Add `GiniBottomSheetAccessibilityRestorable` conformance to the coordinator.**
Same file, append at the end (or before the existing `// MARK: - Document Metadata creation` extension, around line 1225):

```swift
// MARK: - Accessibility restoration
extension GiniBankNetworkingScreenApiCoordinator: GiniBottomSheetAccessibilityRestorable {
    var presenterViewController: UIViewController { screenAPINavigationController }
}
```

This is the piece that lets the wired `onDismiss` callbacks call `restoreAccessibilityAfterBottomSheetDismissal()`. The protocol's default `accessibilityFocusTargetViewController` implementation already picks `navigationController.topViewController` for us — no override needed.

---

### Verification

Simulator with VoiceOver on (⌘F5): analyze a credit-note document with both hint flags true → sheet appears, focus lands on the icon/title, swipes stay trapped (already works). Tap **Cancel** — swipe right on the presenter, focus should land on the nav bar or analysis screen underneath. Repeat for **Proceed** and the already-paid warning sheet. Regression-check: Edit Line Item sheet still restores focus (same protocol; edits here are additive).

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
```diff
@@ -1004,6 +1073,24 @@ extension GiniBankNetworkingScreenApiCoordinator: SkontoCoordinatorDelegate {
         documentWarningViewController.presentAsBottomSheet(from: screenAPINavigationController)
```
zladzeyka, line 1193: [Suggestion]: Wire the restore hook so VoiceOver focus returns to the presenter after the already-paid warning sheet dismisses.

**Impact today:** the `presentAsBottomSheet(from:)` call on this line sets `screenAPINavigationController.view.accessibilityElementsHidden = true` on presentation. When the user taps Cancel or Proceed, that flag is never flipped back — the presenter reports zero accessible elements and any screen presented on this nav controller afterwards inherits the invisible container. This wire-up (plus the coordinator conformance + `onDismiss` property described in the review body) closes the leak for the already-paid flow.

```suggestion
        documentWarningViewController.onDismiss = { [weak self] in
            self?.restoreAccessibilityAfterBottomSheetDismissal()
        }

        documentWarningViewController.presentAsBottomSheet(from: screenAPINavigationController)
```

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
```diff
@@ -1004,6 +1073,24 @@ extension GiniBankNetworkingScreenApiCoordinator: SkontoCoordinatorDelegate {
...
+        }, onProceed: { [weak self] in
+            self?.handleSavingPhotos(for: extractionResult)
+            self?.screenAPINavigationController.dismiss(animated: true) {
+                onProceedTapped()
+            }
+        })
+
+        documentWarningViewController.isModalInPresentation = true
+
+        documentWarningViewController.presentAsBottomSheet(from: screenAPINavigationController)
```
zladzeyka, line 1211: [Suggestion]: Same wiring for the credit-note warning sheet — same failure mode. After the user chooses Cancel or Proceed, `screenAPINavigationController.view.accessibilityElementsHidden` stays `true` because `presentAsBottomSheet(from:)` set it on presentation and nothing lifts it. On the Proceed path this is especially bad: `presentTransactionDocsAlert` (line 551) may present a follow-up alert on the same nav controller whose view is already invisible to VoiceOver. Restoration only happens through `restoreAccessibilityAfterBottomSheetDismissal()`, which is why the callback has to be wired here as well.

```suggestion
        documentWarningViewController.isModalInPresentation = true
        documentWarningViewController.onDismiss = { [weak self] in
            self?.restoreAccessibilityAfterBottomSheetDismissal()
        }

        documentWarningViewController.presentAsBottomSheet(from: screenAPINavigationController)
```

BankSDK/GiniBankSDKExample/GiniBankSDKExample/Credentials.plist
```diff
@@ -7,10 +7,14 @@
 	<key>client_password</key>
 	<string>***</string>
 	<key>client_id</key>
-	<string>***</string>
+	<string>gini-mobile-test</string>
```
zladzeyka:
```suggestion
	<string>***</string>
```

zladzeyka:
```suggestion
	<string>***</string>
```

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/CreditNoteWarning/CreditNoteWarningViewController.swift
```diff
@@ -0,0 +1,98 @@
+//
+//  CreditNoteWarningViewController.swift
+//
+//  Copyright © 2025 Gini GmbH. All rights reserved.
```
zladzeyka:
```suggestion
//  Copyright © 2026 Gini GmbH. All rights reserved.
```

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Custom views/InfoBottomSheet/InfoBottomSheetViewController.swift
```diff
@@ -350,12 +354,16 @@ public class InfoBottomSheetViewController: GiniBottomSheetViewController {
...
         var elements: [Any] = isIphoneAndLandscape ? [] : [iconImageView]
-        elements += [
-            headerLabel,
-            descriptionLabel,
-            buttonsViewContainer.primaryButton,
-            buttonsViewContainer.secondaryButton
-        ]
+
+        // Add header and description
+        elements.append(headerLabel)
```
zladzeyka, line 359: [Suggestion]: VoiceOver announces the title twice. On line 343 the icon's `accessibilityLabel` is set to `viewModel.title` with trait `.image`; then the very next element in the reading order (line 359) is `headerLabel` with trait `.header` — same string, different role. Users hear e.g. "Credit Note Detected, image" immediately followed by "Credit Note Detected, heading."

Drop the icon from the reading order (it's decorative reinforcement of the title anyway) so the header is announced once:

```suggestion
        let isIphoneAndLandscape = UIDevice.current.isIphoneAndLandscape
        // Set explicit VoiceOver navigation order — icon is decorative, header carries the announcement
        var elements: [Any] = []

        // Add header and description
        elements.append(headerLabel)
```

Also set `iconImageView.isAccessibilityElement = false` on line 342 (outside this hunk, cannot inline). Alternatively, give the icon a distinct role-descriptive label (e.g. `"Warning"`) via the content view model rather than duplicating the title.

GiniComponents/Utilities/GiniUtilites/Sources/GiniUtilites/Protocols/BottomSheet/GiniBottomSheetAccessibilityRestorable.swift
```diff
@@ -0,0 +1,69 @@
...
+     bottomSheetVC.onDismiss = { [weak self] in
+     self?.restoreAccessibilityAfterBottomSheetDismissal()
+     }
+     ```
+
+     - Note: A small delay is used to ensure the dismissal animation completes
+     before VoiceOver attempts to refocus.
+     */
+    public func restoreAccessibilityAfterBottomSheetDismissal() {
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) { [weak self] in
```
zladzeyka, line 59: [Suggestion]: The hard-coded `deadline: .now() + 0.1` is racy. `UISheetPresentationController` dismissal animates ~300 ms; on slower devices or with Reduce Motion off, this timer can fire **before** the animation completes, and VoiceOver posts `.screenChanged` while the sheet is still on-screen — focus lands on wrong content or nothing.

Hook the presenter's transition coordinator instead, which fires exactly when the animation completes:

```suggestion
    public func restoreAccessibilityAfterBottomSheetDismissal() {
        let restore = { [weak self] in
            guard let self else { return }

            // Reset accessibility on presenter's view
            self.presenterViewController.view.accessibilityElementsHidden = false

            // If presenter is a navigation controller, also reset the navigation bar
            if let navigationController = self.presenterViewController as? UINavigationController {
                navigationController.navigationBar.accessibilityElementsHidden = false
            }

            // Notify VoiceOver to refocus on the target view controller
            UIAccessibility.post(notification: .screenChanged,
                                 argument: self.accessibilityFocusTargetViewController?.view)
        }

        if let coordinator = presenterViewController.transitionCoordinator {
            coordinator.animate(alongsideTransition: nil) { _ in restore() }
        } else {
            DispatchQueue.main.async(execute: restore)
        }
```

BankSDK/GiniBankSDKExample/Tests/IntegrationTests/TransferSummary/BaseIntegrationTest+CXTransferSummary.swift
```diff
@@ -82,15 +82,19 @@ extension BaseIntegrationTest {
         }
 
         DispatchQueue.main.asyncAfter(deadline: .now() + 10) {
```
zladzeyka, line 84: [Suggestion]: Bare `asyncAfter(deadline: .now() + 10)` reintroduces the fixed-sleep pattern the transfer-summary path just moved away from (see the polling-based `pollUpdatedExtractions` helper used in `BaseIntegrationTest+TransferSummary.swift`). On fast networks this wastes 10 s per run; on slow ones it can still race the backend. Reuse the polling helper — it exits as soon as the expected feedback is visible and gives up cleanly with a diagnostic on exhaustion.

```suggestion
        pollUpdatedExtractions(for: document, expect: expect) { updatedResult in
            DispatchQueue.main.async {
                guard !self.isTestFinished else { return }
                switch updatedResult {
                case .success:
                    /// Backend accepted the `crossBorderPayment` feedback — document is still accessible.
                    GiniBankConfiguration.shared.cleanup()
                    XCTAssertNil(GiniBankConfiguration.shared.documentService)
                    expect.fulfill()
                case let .failure(error):
                    XCTFail("CX transfer summary feedback was rejected by the backend: \(error)")
                }
            }
        }
```

(Adjust the helper name / signature to whatever the transfer-summary side is actually using — the point is: no fixed 10 s sleep.)

BankSDK/GiniBankSDK/Tests/GiniBankSDKTests/GiniBankConfigurationTests.swift
```diff
@@ -8,294 +8,367 @@ import Testing
...
+
+         This acts as a safeguard to ensure configuration changes never go untested.
+         */
+        @Test("GiniBankConfiguration must not change stored properties without updating tests")
+        func configurationPropertyCount() {
+            let mirror = Mirror(reflecting: GiniBankConfiguration())
+
+            // NOTE: Update this value whenever you add or remove a stored property.
+            let propertyCount = mirror.children.count
+            let expectedCount = 67 // Current number of stored properties in GiniBankConfiguration
```
zladzeyka: [Suggestion]: The hard-coded `expectedCount = 67` mirror check fires on **every** stored-property change — feature-flag additions, renames, unrelated refactors — and when it fires it doesn't tell the reviewer what changed. A stronger form asserts on the *set of property names* the suite knows about; a mismatch then names the added or removed property, and the fix is a one-line diff to the expected set.

```suggestion
            let mirror = Mirror(reflecting: GiniBankConfiguration())
            let actual = Set(mirror.children.compactMap { $0.label })

            // Add to / remove from this set when a stored property is added / removed.
            let expected: Set<String> = [
                "bottomNavigationBarEnabled", "multipageEnabled", "qrCodeScanningEnabled",
                "onlyQRCodeScanningEnabled", "flashToggleEnabled", "flashOnByDefault",
                "onboardingShowAtLaunch", "onboardingShowAtFirstLaunch", "openWithEnabled",
                "returnAssistantEnabled", "enableReturnReasons", "skontoEnabled",
                "transactionDocsEnabled", "alreadyPaidHintEnabled", "savePhotosLocallyEnabled",
                "paymentDueHintEnabled", "paymentDueHintThresholdDays", "creditNoteHintEnabled",
                // … fill in the remaining names
            ]
            #expect(actual == expected,
                    "GiniBankConfiguration stored properties drifted from the suite's expectation. Added: \(actual.subtracting(expected)). Removed: \(expected.subtracting(actual)).")
```

The `Added: / Removed:` diagnostic in the failure message is the win — the current message just says "a new property was added" without naming it.

zladzeyka (commented):
## Non-blocking review — quality nits worth surfacing

Four inline `[Suggestion]:` comments below (all inside diff hunks, one-click applicable). Three more items are outside any hunk on this PR, so they can't render as inline suggestions — listed here with `[Suggestion]:` + file:line + copy-pasteable snippets.

---

**[Suggestion]: Dead `import Combine`** in `CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/DocumentPaidWarning/DocumentMarkedAsPaidViewController.swift:8` — no Combine usage remains after the refactor. Drop the import.

```swift
// current:
import UIKit
import Combine   // ← unused, remove

// after:
import UIKit
```

---

**[Suggestion]: `adjustsFontSizeToFitWidth` defeats Dynamic Type** in `CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Custom views/InfoBottomSheet/InfoBottomSheetViewController.swift:65`. Combined with `adjustsFontForContentSizeCategory = true` (line 64) and `numberOfLines = 0` (line 60), the header will shrink localized titles at AX text sizes rather than wrapping. WCAG 2.2 1.4.4 / BFSG expects 200% scaling without loss of content — and the sheet already switches to `.large()` at `isFontSizeAtLeastAccessibilityMedium`, so wrapping is safe.

```swift
// remove this line:
label.adjustsFontSizeToFitWidth = true
```

---

**[Suggestion]: Lying doc comment on a no-op override** in `BankSDK/GiniBankSDKExample/Tests/IntegrationTests/TransferSummary/CXTransferSummaryIntegrationTest.swift:45-48`. The `handleUploadSuccess` override just calls `super`, but the preceding doc comment (lines 40-44) claims it flips the product tag to `cxExtractions` before `startAnalysis()`. Either restore the tag flip (matches what the doc says) or delete the override entirely — today the comment lies about the behavior.

```swift
// today:
override func handleUploadSuccess(captureDocument: GiniCaptureDocument,
                                  delegate: GiniCaptureResultsDelegate) {
    super.handleUploadSuccess(captureDocument: captureDocument, delegate: delegate)
}
```

zladzeyka (changes requested):
Thanks @ValentinaIancu-Gini 
I left a couple of suggestions. The one important about accessibility focus restoration when we dismiss the bottom sheets tapping on the cancel transfer or other cancel actions

BankSDK/GiniBankSDK/Tests/GiniBankSDKTests/GiniBankConfigurationTests.swift
ValentinaIancu-Gini: I think it is complicating more things. I prefer to update one number instead of keeping in mind to duplicate the new property in the set

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Custom views/InfoBottomSheet/InfoBottomSheetViewController.swift
ValentinaIancu-Gini: this is intended.

GiniComponents/Utilities/GiniUtilites/Sources/GiniUtilites/Protocols/BottomSheet/GiniBottomSheetAccessibilityRestorable.swift
ValentinaIancu-Gini: I see no issue with this. I will not change it now.

BankSDK/GiniBankSDKExample/Tests/IntegrationTests/TransferSummary/BaseIntegrationTest+CXTransferSummary.swift
ValentinaIancu-Gini: @zladzeyka I don't understand why I need to do this change. Can we skip it for now and come back when it becomes a problem? Until now tests are passing successfully so I see no issues in skipping this fix.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/DocumentPaidWarning/DocumentMarkedAsPaidViewController.swift
```diff
@@ -5,14 +5,14 @@
...
 import UIKit
-import Combine
 
 private struct DocumentMarkedAsPaidContentViewModel: InfoBottomSheetViewModel {
     var image: UIImage? = UIImageNamedPreferred(named: "infoMessageIcon")
     var imageTintColor: UIColor? = GiniColor(light: .GiniCapture.warning2,
                                              dark: .GiniCapture.warning2).uiColor()
     var title: String = DocumentMarkedAsPaidViewController.Strings.title
     var description: String = DocumentMarkedAsPaidViewController.Strings.description
+    var imageBackgroundColor: UIColor? = DocumentMarkedAsPaidViewController.Colors.imageBGColor
```
zladzeyka, line 15: [Clarification]: why we didn't use `GiniColor` here like we've done for `imageTintColor`

.github/workflows/bank-sdk.check.yml
```diff
@@ -8,6 +8,14 @@ on:
       - 'BankSDK/**'
       - 'CaptureSDK/**'
       - 'BankAPILibrary/**'
```
zladzeyka, line 10: [Clarification]: Do we need to add GiniUtilites there?

.github/workflows/bank-sdk.check.yml
```diff
@@ -8,6 +8,14 @@ on:
...
+  # Analyze the main branch in SonarCloud after every merge so PR quality gates
+  # have a new-code baseline to compare against.
+  push:
+    branches: [main]
+    paths:
+      - 'BankSDK/**'
+      - 'CaptureSDK/**'
+      - 'BankAPILibrary/**'
```
zladzeyka, line 19: [Clarification]: Do we need to add GiniUtilites there?

.github/workflows/health-sdk.check.yml
```diff
@@ -7,6 +7,13 @@ on:
     paths:
       - 'HealthSDK/**'
       - 'HealthAPILibrary/**'
+  # Analyze the main branch in SonarCloud after every merge so PR quality gates
+  # have a new-code baseline to compare against.
+  push:
+    branches: [main]
+    paths:
+      - 'HealthSDK/**'
+      - 'HealthAPILibrary/**'
```
zladzeyka, line 18: [Clarification]: Do we need to add GiniUtilites and GiniInternalPaymentSDK there?

ValentinaIancu-Gini: Good question. I will check. also for comments above.

CaptureSDK/GiniCaptureSDK/Sources/GiniCaptureSDK/Core/Screens/DocumentPaidWarning/DocumentMarkedAsPaidViewController.swift
zladzeyka: Decision was keeping like its

BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
zladzeyka: @ValentinaIancu-Gini confirmed that there is no issues

zladzeyka (approved):

zladzeyka (approved):

PR #984, opened 2025-09-25
ci: Fastlane - add a new lane to register new devices from a devices …

ValentinaIancu-Gini:
## Pull Request Description

[PP-1834](https://ginis.atlassian.net/browse/PP-1834)
This PR adds a new Fastlane lane to register Apple devices from a file, streamlining the device registration process for iOS development. The change introduces automation for bulk device registration using a structured text file format.

- Adds a register_new_devices lane to the Fastfile with comprehensive documentation
- Creates a devices.txt file with example device entries in tab-separated format
- Includes support for both iOS and macOS devices with proper UDID formats

[PP-1834]: https://ginis.atlassian.net/browse/PP-1834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1245, opened 2026-08-20
ai: Align gini-release skill with Mobile Release Process

zladzeyka:
## Pull Request Description

Refines the `/gini-release` skill introduced in #1232 to follow the iOS steps of the [Mobile Release Process](https://ginis.atlassian.net/wiki/spaces/PLMO/pages/83689511/Mobile+Release+Process) Confluence page and to address review feedback from Valentina, Copilot, Mozhgan and Qazi across the review cycle.

Base branch is `create-release-skill` so the diff shows only the refinements on top of #1232. Internal tooling — no Jira ticket.

### Structural change: split into two phases

The skill was previously a single `SKILL.md`. It is now three files:

- **`SKILL.md`** — router. Reads the invocation argument (`rc` / `release` / no arg) and hands off to the matching phase file.
- **`phase-1-rc.md`** — Jira-only work: confirm the release branch, dispatch the XCFramework build against it, verify Jira releases + `x.x` placeholder versions exist, create the RC ticket with the QA build (Firebase link + QR code), put the ticket in the active sprint and hand it to QA.
- **`phase-2-tag-and-publish.md`** — git work: version bumps on the release branch, bump PR, `create_release_tags` from the release branch, XCFrameworks, draft GitHub releases with a hard user-review gate, publish, CocoaPods podspec, Jira closeout, TestFlight upload, merge back to `main`, `#mobile-releases` announcement.

Split mirrors [Mozhgan's Android PR #984](https://github.com/gini/gini-mobile-android/pull/984). Phase 1 stays close to the Android version; phase 2 is iOS-specific.

### Phase 1 (RC)

- **Verify Jira Releases exist**, don't batch-create — releases are normally created at release planning; this step is a safety check with a create-only-if-missing fallback. `fixVersion` assignment stays a planning decision, not something the skill batch-runs.
- `<major>.x.x` placeholder versions per product owning notes.
- **Do-not-touch list**: `iOS Unknown Fix Version` (PP), `Android Unknown Fix Version` (PP sibling), `BAC - placeholder bugs`, `CVIE placeholder bugs`.
- Release-branch XCFramework build dispatched before the RC ticket so the RC has a build to attach.
- RC ticket attaches Firebase console link, tester-app link, QR code, XCFramework workflow run link.
- Sprint assignment + handoff via `transitionJiraIssue` (`In Progress` → `Waiting for QA`).

### Phase 2 (bump, tag, publish)

- **Step 1 hard gate** with a Jira pre-check: reads the RC ticket's `status.name` via the Atlassian connector and surfaces it in the confirmation prompt. Status `In QA` / `Done` → *"confirm sign-off"*; anything else → *"still `<status>` — sure you want to proceed?"* The user still confirms explicitly — status is a hint, not the gate.
- Bumps run on the release branch (not `main`) so a failing release can be fixed there first.
- Bump PR opened from the release branch; approval on it is the second sub-gate before pushing tags.
- `create_release_tags` runs from the release branch.
- Post-tag flow: build XCFrameworks → draft GitHub releases in mono + individual release repos → hard user-review gate → publish → CocoaPods podspec push for `GiniBankSDK` → Jira releases closed out.
- **New step 12 — TestFlight upload.** Dispatches [`bank-sdk.publish.example.app.testflight.yml`](https://github.com/gini/gini-mobile-ios/actions/workflows/bank-sdk.publish.example.app.testflight.yml) / [`health-sdk.publish.example.app.testflight.yml`](https://github.com/gini/gini-mobile-ios/actions/workflows/health-sdk.publish.example.app.testflight.yml) via `workflow_dispatch` before the Slack announcement. The quarterly cron on those workflows rarely lines up with a release, so the announced build was often stale — this fixes it.
- **Step 13 — Slack announcement template** derived from the Bank SDK 4.5.0 post. Includes `<platform>` slot so Android's paired skill can reuse the same shape, per-feature bullets with Confluence deep links, release-notes URLs, TestFlight join + QR code.

### Other cleanups addressed via review

- Never release from `main` — always a `release/<theme>` branch. `create_release_tags` gated on that.
- `Installation.md` list corrected to three files including `GiniHealthSDK`.
- Version-list example expanded to both chains; `GiniUtilites` misspelling reframed (tag must mirror the package name; fix in a future major).
- `create_release_tags` gotcha: delete unpushed local tag with `git tag -d "<Package>;<version>"` (quoted — `;` is a shell separator) before rerunning.
- Beta-tag caveat: only `GiniBankAPILibrary`, `GiniCaptureSDK`, `GiniBankSDK` workflows match `-betaNN`.
- Podspec section fleshed out with a placeholder legend + a real example.
- Cross-references disambiguated (e.g. "phase 1's step 1 packages table" instead of the ambiguous "phase 1's table").
- Handoff wording on phase-1's final "next step" fixed — QA sign-off is the only pre-condition for starting phase 2 (the bump PR gate happens inside phase 2, not before).

## Notes for Reviewers

Documentation-only — no code, no simulator/device runs required.

Please verify by reading each file end-to-end:

- [ ] `SKILL.md` router table (phase scope + invocation) matches the two phase files.
- [ ] `phase-1-rc.md` step 5's verify-not-create framing is right — releases are normally created at release planning; this step is a safety check with a create-only-if-missing fallback.
- [ ] `phase-1-rc.md` step 6's do-not-touch list is complete: `iOS Unknown Fix Version`, `Android Unknown Fix Version`, `BAC - placeholder bugs`, `CVIE placeholder bugs`.
- [ ] `phase-2-tag-and-publish.md` step 1's Jira status pre-check uses the right statuses (`In QA` / `Done` → sign-off; anything else → warn).
- [ ] `phase-2-tag-and-publish.md` step 12's `gh workflow run` invocations are right for both TestFlight workflows.
- [ ] `phase-2-tag-and-publish.md` step 13's Slack template covers what the Bank SDK 4.5.0 announcement had; the `<platform>` slot works for Android's paired skill.
- [ ] Cross-references between the two phase files are unambiguous, and phase-1's final "next step" line no longer requires the bump PR to be approved before running phase 2.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,323 @@ The seven releasable packages, their version files and their release repos:
...
+From the repo root:
+
+```bash
+bundle exec fastlane create_release_tags
+```
+
+Run this **manually in the terminal** — the lane needs an interactive prompt (`! bundle exec fastlane create_release_tags`).
+
+- The lane scans every `**/*Version.swift`, compares each package's version against its latest `<Package>;<version>` release tag, creates a local tag for every package that differs, and prompts **"Push release tag?"** per package.
+- **The lane creates the local tag before asking to push.** If a push prompt is declined, the local tag stays and the lane sees the package as up-to-date on the next run. Delete the unpushed local tag before rerunning: `git tag -d <Package>;<version>`.
```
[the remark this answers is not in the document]
zladzeyka: Will keep like it's for now

.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,323 @@ The seven releasable packages, their version files and their release repos:
...
+Confirm `git log -1` points at the release merge commit — `create_release_tags` tags the current `HEAD`, so an out-of-date `main` tags the wrong commit.
+
+### 5b. Verify the XCFramework build prerequisite
+
+Verify the XCFramework build workflows have completed successfully on the release branch that just merged:
+
+- `bank-sdk.build.xcframeworks` — if the release includes `GiniBankSDK`.
+- `health-sdk.build.xcframeworks` — if the release includes `GiniHealthSDK`.
+
+Do not proceed before the applicable workflows are green.
```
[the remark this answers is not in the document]
zladzeyka: Not correct, we need to make sure that workflow for the xcframeworks has successful run

.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,323 @@ The seven releasable packages, their version files and their release repos:
...
+XCFrameworks are built for both `GiniBankSDK` and `GiniHealthSDK`. Skip either sub-step if that SDK isn't in this release; jump to step 7 if neither is.
+
+Start the XCFramework builds now so the archives are ready by the time the release drafts are reviewed. On `main`, on the commit that already carries the release tags, add one tag per SDK being released:
+
+```
+GiniBankSDK;<X.Y.Z>;xcframeworks
+GiniHealthSDK;<X.Y.Z>;xcframeworks
+```
+
+Each tag triggers the matching workflow (`bank-sdk.build.xcframeworks` / `health-sdk.build.xcframeworks`).
```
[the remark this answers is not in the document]
zladzeyka: addressed

.claude/skills/gini-release/SKILL.md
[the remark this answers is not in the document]
zladzeyka: Fixed in 2e2a3472c — the `git tag -d` example now quotes the argument:

```bash
git tag -d "<Package>;<version>"
```

That prevents the shell from splitting on `;`. Spot-check: lines 271 / 277 / 307 already quote the tag (`"<Package>;<X.Y.Z>"`), so only line 225 needed this fix.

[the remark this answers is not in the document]
zladzeyka: Went ahead and quoted it in 2e2a3472c. As-written, `git tag -d GiniBankSDK;4.4.0` runs `git tag -d GiniBankSDK` and then tries to execute `4.4.0` — the actual local release tag is not deleted, so the hint does the opposite of what it claims. One-character fix (add quotes), no runtime downside. Happy to revert if you prefer to keep the original phrasing.

.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,323 @@ The seven releasable packages, their version files and their release repos:
...
+https://ginis.atlassian.net/projects/<KEY>?selectedItem=com.atlassian.jira.jira-projects-plugin:release-page&status=all
+```
+
+`status=all` matters — the default filter hides released versions. Match on name; permanent `UNRELEASED` versions are kept for parking tickets.
+
+In the `Create release` dialog: fill Release name and Description, and **clear the prefilled Release date** (defaults to today, wrong for an unshipped version — click field, `cmd+a`, `Backspace`, dismiss picker). Reload the page after creating — the table doesn't refresh, so success looks like failure. Read each version's numeric id off its table link (`/projects/<KEY>/versions/<id>/tab/…`).
+
+Once versions exist, assign them as `fixVersions` on all work tickets in the release using `editJiraIssue` (names work here). Without fix versions the release report is empty.
+
+Add **release notes in markdown** to each release description. These notes are reused verbatim on GitHub releases (step 6) — copy from the previous release and update.
```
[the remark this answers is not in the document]
zladzeyka: Fixed in 2e2a3472c — line 93 now says step 7 (drafts), not step 6 (XCFramework builds). Spot-checked the other locations Copilot cited (154 / 261 / 356) — none of them contained a wrong step reference, so no changes needed there.

.claude/skills/gini-release/SKILL.md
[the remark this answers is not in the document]
zladzeyka: Reworded in 2e2a3472c. 5b now explicitly targets the **branch-triggered** XCFramework workflow run for the release branch that just merged into `main`, and a parenthetical clarifies that step 6 later triggers a separate **tag-triggered** run of the same workflow to produce the archives attached to the drafts. Contradiction resolved while keeping the safeguard from @zladzeyka's follow-up.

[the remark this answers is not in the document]
zladzeyka: Agreed — the requirement is kept. Reworded 5b in 2e2a3472c so it's explicit that we verify the **branch-triggered** run for the release branch is green before pushing release tags, and noted that step 6 later triggers a separate **tag-triggered** run to produce the archives for the drafts. Please let me know if the branch-vs-tag split matches how you actually verify.

.claude/skills/gini-release/SKILL.md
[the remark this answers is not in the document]
zladzeyka: Addressed across two commits:

- **2e2a3472c** — step 6 now shows explicit `git tag` + `git push origin` commands for both `GiniBankSDK;<X.Y.Z>;xcframeworks` and `GiniHealthSDK;<X.Y.Z>;xcframeworks`, and notes that only pushing triggers the workflow.
- **3aa33ee9a** — replaced "verify the runs started" with a hard "**wait until both runs finish successfully**" gate. Step 7 attaches the generated archives, so a merely-started run isn't enough.

The second location Copilot cited (line 263 / step 7 header) reads "For each pushed `<Package>;<X.Y.Z>` tag …" — this is now grounded because step 5f's fastlane lane and step 6's explicit push commands make it unambiguous which tags have been pushed.

MozhganPeivandianSharbaf (approved):
Looks good, Thank you 👍

.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,326 @@ The seven releasable packages, their version files and their release repos:
...
+
+Assign the RC ticket to the QA engineer.
+
+Report the RC ticket key — it goes in the final release commit.
+
+## 3. Create the release candidate branch
+
+**Always release from a `release/<theme>` branch, never from `main`.** Feature PRs for the release merge into this branch; the theme usually matches the Jira fix versions suffix (`release/qr-code-improvements`, `release/liquid_glass_bank_sdk`, `release/bank_sdk_release_4.2.2`). Ask which branch this release ships from — do not assume.
+
+Confirm the branch is up to date with `main` before proceeding.
```
ValentinaIancu-Gini: [Suggestion] I think this should be first thing, before creating the rc with the build. We should trigger the build out of this branch.

.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,326 @@ The seven releasable packages, their version files and their release repos:
...
+
+## 4. Wait for QA — hard gate
```
ValentinaIancu-Gini: [Suggestion] Mozhgan already split this in 2 parts. First she did it in Android. We should align this.

.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,326 @@ The seven releasable packages, their version files and their release repos:
...
+
+**GiniHealthSDK** — from the `health-sdk.build.xcframeworks` run:
+
+1. Download the generated artifact (uploaded as `GiniHealthSDKFramework`) and unzip.
+2. Create an archive `GiniHealthSDK_<X.Y.Z>_XCFrameworks` containing only `GiniHealthSDK`, `GiniHealthAPILibrary`, `GiniInternalPaymentSDK`, `GiniUtilites`.
+3. Keep it locally — it will be attached in step 7.
+
+See the [GiniBankSDK 3.0.0 release](https://github.com/gini/gini-mobile-ios/releases/tag/GiniBankSDK%3B3.0.0) for what the final attached archive looks like.
+
+## 7. Draft the GitHub releases
```
ValentinaIancu-Gini: [Suggestion] @zladzeyka we should provide a model of how we want to have release notes for each repo.

.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,326 @@ The seven releasable packages, their version files and their release repos:
...
+- Mono repo `gini/gini-mobile-ios` — every `<Package>;<version>` tag gets a draft here.
+- Individual release repo, if the package has one: `gini/bank-api-library-ios`, `gini/capture-sdk-ios`, `gini/bank-sdk-ios`, `gini/health-api-library-ios`, `gini/internal-payment-sdk-ios`, `gini/utilites-ios`, `gini/health-sdk-ios`.
+
+**Use the previous release of the same SDK as the template.** Fetch its notes to mirror title formatting and section structure:
+
+```bash
+gh release view "<Package>;<previous-X.Y.Z>" --repo <repo> --json name,body
+```
+
+Fill the draft body from the Jira release description markdown, following the previous release's shape. Then create the draft:
```
ValentinaIancu-Gini: [Clarification] I don't know what is this. Can you please give me an example?

.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,326 @@ The seven releasable packages, their version files and their release repos:
...
+- **Description** — three sections:
+  1. `**Issue Summary**` — the line "Here is the list of tickets for the release." then one `https://ginis.atlassian.net/browse/<TICKET>` link per ticket in the release. Find them with `fixVersion in ("<version-name>", …)`.
+  2. `**Listed Releases**` — one Jira release-report link per fix version: `https://ginis.atlassian.net/projects/<KEY>/versions/<id>/tab/release-report-all-issues`
+  3. `**Attachments**:` — "Build for testing can be found here:" plus the Firebase App Distribution console link and the tester-app link for the example app. **Ask the user** — these come from the CI/Firebase build.
+
+Set the same `fixVersions` on the RC ticket. Put it in the **active sprint**: fetch a ticket in `sprint in openSprints()` with `expand: names`, find the `Sprint` custom field, set the entry whose `state` is `active`.
+
+**Drive status via `transitionJiraIssue`** — resolve transitions per ticket by name, case-insensitively via `getTransitionsForJiraIssue`. Ids collide across projects and can silently move a ticket to `Cancelled`.
+
+Assign the RC ticket to the QA engineer.
```
ValentinaIancu-Gini: [Suggestion] We can ask AI now to generate Qrcode to attache it to the ticket. I did it and worked perfectly. Please add it somewhere here.

.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,326 @@ The seven releasable packages, their version files and their release repos:
...
+### 5a. Merge and switch to main
+
+Merge the release PR into `main` (or confirm it has already merged). Then switch explicitly:
+
+```bash
+git checkout main
+git pull --ff-only
+```
+
+Confirm `git log -1` points at the release merge commit — `create_release_tags` tags the current `HEAD`, so an out-of-date `main` tags the wrong commit.
```
ValentinaIancu-Gini: [Clarification] @zladzeyka we normally do this after we release. I would keep it like that.

.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,326 @@ The seven releasable packages, their version files and their release repos:
...
+https://ginis.atlassian.net/projects/<KEY>?selectedItem=com.atlassian.jira.jira-projects-plugin:release-page&status=all
+```
+
+`status=all` matters — the default filter hides released versions. Match on name; permanent `UNRELEASED` versions are kept for parking tickets.
+
+In the `Create release` dialog: fill Release name and Description, and **clear the prefilled Release date** (defaults to today, wrong for an unshipped version — click field, `cmd+a`, `Backspace`, dismiss picker). Reload the page after creating — the table doesn't refresh, so success looks like failure. Read each version's numeric id off its table link (`/projects/<KEY>/versions/<id>/tab/…`).
+
+Once versions exist, assign them as `fixVersions` on all work tickets in the release using `editJiraIssue` (names work here). Without fix versions the release report is empty.
+
+Add **release notes in markdown** to each release description. These notes are reused verbatim on GitHub releases (step 7) — copy from the previous release and update.
```
ValentinaIancu-Gini: [Clarification] I don't see anything for creating the placeholders for the next release in Jira. I would suggest to make it as a step here.

ValentinaIancu-Gini (commented):
@zladzeyka I reviewed this skill. Please check my comments, thank you!

.claude/skills/gini-release/SKILL.md
zladzeyka: Thanks — before I split step 4, could you link Mozhgan's Android skill (or paste the two parts here)? Want to mirror her split rather than guess at where the seam should go.

.claude/skills/gini-release/SKILL.md
zladzeyka: Could you clarify which specific action should move to "after we release"? A few readings:

1. The whole version-bump-and-tag flow of step 5 — happens after publish rather than after QA sign-off
2. Just this specific `git log -1` verification checkpoint
3. Something else

If (1), that's a bigger reordering (versioning happens before public tags → releases → notes) — I'd rather confirm before touching. Happy to go whichever way you had in mind.

.claude/skills/gini-release/SKILL.md
zladzeyka: Done in e7582bfeb — new step 3b dispatches the XCFramework build on the release branch pre-QA; old step 5b removed.

.claude/skills/gini-release/SKILL.md
zladzeyka: Done in e7582bfeb — added a callout in step 7 pointing at the planned /gini-release-notes companion skill for per-repo templates.

.claude/skills/gini-release/SKILL.md
zladzeyka: Done in 3611df47f — folded into the V3 delegation; the worked Jira → GitHub example lives in /gini-release-notes.

.claude/skills/gini-release/SKILL.md
zladzeyka: Done in e7582bfeb — Attachments now include a QR code of the tester-app link so QA can scan and install without typing.

.claude/skills/gini-release/SKILL.md
zladzeyka: Done in e7582bfeb — added a note at the end of section 2a clarifying that next-cycle placeholders are created in step 11, not now.

.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,340 @@ The seven releasable packages, their version files and their release repos:
...
+
+Set the same `fixVersions` on the RC ticket. Put it in the **active sprint**: fetch a ticket in `sprint in openSprints()` with `expand: names`, find the `Sprint` custom field, set the entry whose `state` is `active`.
+
+**Drive status via `transitionJiraIssue`** — resolve transitions per ticket by name, case-insensitively via `getTransitionsForJiraIssue`. Ids collide across projects and can silently move a ticket to `Cancelled`.
+
+Assign the RC ticket to the QA engineer.
+
+Report the RC ticket key — it goes in the final release commit.
+
+## 3. Create the release candidate branch
```
ValentinaIancu-Gini: [Clarification] @zladzeyka  I still think that this step should be the first step before creating even the RC ticket. I left a comment below but it wasn't understood correctly

.claude/skills/gini-release/SKILL.md
ValentinaIancu-Gini: https://github.com/gini/gini-mobile-android/pull/984

.claude/skills/gini-release/SKILL.md
ValentinaIancu-Gini: Normally, we do `create_release_tags` from the release branch, not from main. In case the release is not working or something is crashing to have a chance to fix it in the release branch not on main.

.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,340 @@ The seven releasable packages, their version files and their release repos:
...
+### 5b. Bump versions in dependency order
+
+For each package in `RELEASE-ORDER.md` order, edit **up to four** places. Missing any of the last three is how iOS releases break silently:
+
+1. **The version file** — `public let <Package>Version = "<x.y.z>"` in the path from step 1's table.
+2. **Every dependent's `Package-release.swift`** — bump the `.exact("<x.y.z>")` pin. PR checks only resolve `Package.swift`, not `Package-release.swift`, so a missed pin lands silently:
 
    ```bash
    grep -rn '\.exact(' --include='Package-release.swift' .
    ```
```
ValentinaIancu-Gini: As I said this should not happen on main branch.

.claude/skills/gini-release/SKILL.md
ValentinaIancu-Gini: So should I understand that `following the previous release's shape` is fetching the previous release notes? if yes, that's nice. Smart.

.claude/skills/gini-release/SKILL.md
```diff
@@ -45,254 +42,340 @@ The seven releasable packages, their version files and their release repos:
...
+
+## 4. Wait for QA — hard gate
+
+Stop. The RC ticket is with QA. Do not touch versions or tags. Ask the user to confirm the outcome; never infer.
+
+- **QA passes:** proceed to step 5.
+- **QA fails, showstopper:** release postponed. Add details to the RC ticket and stop.
+- **QA fails, minor:** release not postponed. Create bug tickets with the release version set as `affectedVersion` and proceed once fixed.
+
+## 5. Merge, bump versions, create tags
```
ValentinaIancu-Gini: [Request] This is not true. 
What is in this skill, starting with section 5 needs to be updated.
It should be the **release branch,** **not the main branch.** Please update this. Also introduce a new step for merging the release branch on main when everything is done.

ValentinaIancu-Gini (commented):
@zladzeyka I checked one more time and left a couple of comments. Thank you!

.claude/skills/gini-release/SKILL.md
zladzeyka: Yes correctly, fetching the previous release - I decided this the best option since we might change the format in the future

.claude/skills/gini-release/SKILL.md
zladzeyka: Done in the split — phase 1 now runs step 2 (confirm release branch) and step 3 (dispatch `Build BankSDK/HealthSDK XCFrameworks` on that branch) **before** step 7 (create the RC ticket). The RC is only opened once the branch build is green, and its Attachments section links the successful run so QA has a reference build for the frameworks side. See `phase-1-rc.md` steps 2, 3 and 7.

.claude/skills/gini-release/SKILL.md
zladzeyka: Split done in the new commit, following Mozhgan's shape from [PR #984](https://github.com/gini/gini-mobile-android/pull/984). `SKILL.md` is now a router pointing at `phase-1-rc.md` (Jira-only prep, close to the Android phase 1) and `phase-2-tag-and-publish.md` (iOS-specific: bumps, `create_release_tags`, XCFrameworks, GH release drafts, podspec, Jira, Slack). Invocation is `/gini-release rc` for phase 1 and `/gini-release` for phase 2.

.claude/skills/gini-release/SKILL.md
zladzeyka: Fixed in phase 2. The bumps now happen on the release branch (step 2) and `create_release_tags` runs from that same branch (step 5), not from `main`. The merge into `main` is the last step of phase 2 (step 12), after the release ships — so if a release breaks, the fix goes onto the release branch first.

zladzeyka:
@ValentinaIancu-Gini pushed the changes for your remaining comments:

- **Split into two phases** (your comment on line 150, aligned with [Mozhgan's #984](https://github.com/gini/gini-mobile-android/pull/984)) — `SKILL.md` is now a router; the flow lives in `phase-1-rc.md` (Jira-only, close to Mozhgan's Android version) and `phase-2-tag-and-publish.md` (iOS-specific: bumps, `create_release_tags`, XCFrameworks, GH release drafts, podspec, Jira, Slack). Invocation is `/gini-release rc` for phase 1, `/gini-release` for phase 2.
- **Release-branch build first, before the RC ticket** (line 125/129) — phase 1 now confirms the branch → dispatches the XCFramework build → creates Jira Releases + placeholders → **then** creates the RC ticket with the build already attached. See `phase-1-rc.md` steps 2, 3 and 7.
- **`x.x` placeholder versions as a real step** (line 93) — phase 1 step 6 is dedicated to them, with the "create only if missing" rule and the reminder not to touch the per-customer bug-parking versions.
- **`create_release_tags` from the release branch, not `main`** (line 171) — the bumps and the lane both run on the release branch in phase 2 (steps 2 and 5); merge back to `main` is step 12.

The QR-code attachment (line 121) and the release-notes-model + example delegation to `/gini-release-notes` (lines 274, 291) from the earlier round are preserved, now in the phase-1 file.

Ready for another look when you have time — thank you!

.claude/skills/gini-release/phase-1-rc.md
```diff
@@ -0,0 +1,203 @@
...
+|---|---|
+| **PP** (Photopayment, bank side) | `GiniBankAPILibrary`, `GiniCaptureSDK`, `GiniBankSDK` |
+| **HEAL** (Insurance, health side) | `GiniHealthAPILibrary`, `GiniHealthSDK` |
+
+`GiniUtilites` and `GiniInternalPaymentSDK` still get their own git tag + GitHub release in phase 2, but **no Jira release** — their notes live in the parent SDK's release. Their release-notes source is defined in the (planned) `/gini-release-notes` skill; until it exists, follow the previous release of the same package as the template.
+
+Don't take this table on faith if the release looks unusual: the project's Releases page (step 5) is the source of truth for which products have versions.
+
+## 5. Create the Jira Releases (fix versions)
+
```
ValentinaIancu-Gini: [Clarification] I think this is not happening like this. We should already have the jira releases before everything. Maybe we can check if we already have. But the process should be different. Once we already plan a new release we should have already tickets assigned to that Jira release. So I see this as a duplication of stepsl

.claude/skills/gini-release/phase-1-rc.md
```diff
@@ -0,0 +1,203 @@
...
+```
+https://ginis.atlassian.net/projects/<KEY>?selectedItem=com.atlassian.jira.jira-projects-plugin:release-page&status=all
+```
+
+`status=all` matters — the default filter hides released versions. Match on **name**; permanent `UNRELEASED` placeholders are kept for parking tickets (step 6), so never take the last row.
+
+In the `Create release` dialog: fill Release name and Description, and **clear the prefilled Release date** (defaults to today, wrong for an unshipped version — click field, `cmd+a`, `Backspace`, dismiss picker; setting an empty string via `form_input` does not work). After saving, **reload the page** — the table doesn't refresh, so success looks like failure. Don't click `Create release` twice either; the second click closes the dialog. Read each version's numeric id off its table link (`/projects/<KEY>/versions/<id>/tab/…`).
+
+Once versions exist, assign them as `fixVersions` on all work tickets in the release using `editJiraIssue` (names work here). Without fix versions the release report is empty.
```
ValentinaIancu-Gini: [Clarification] How can we do this if we don't have yet tickets attached to a release?

.claude/skills/gini-release/phase-1-rc.md
```diff
@@ -0,0 +1,203 @@
...
+
+e.g. `iOS Gini Bank SDK 4.x.x`, `iOS Gini Capture SDK 4.x.x`, `iOS Gini APILibrary SDK 4.x.x` in PP; `iOS Gini Health API Library 6.x.x`, `iOS Gini Health SDK 6.x.x` in HEAL.
+
+Rules:
+
+- One per product listed in step 4 for the side(s) being released — **not** for `GiniUtilites` or `GiniInternalPaymentSDK`.
+- **Create only if missing.** The placeholder lives for the whole major line and is *not* recreated every release. If `iOS Gini Bank SDK 4.x.x` already exists, leave it alone.
+- `UNRELEASED`, no release date, no description needed. Created in the browser, same dialog and same gotchas as step 5.
+- When a new major line starts (e.g. the first 5.0.0 release), the new placeholder is `5.x.x`; the old `4.x.x` stays until its remaining tickets are moved.
+- **Do not touch** the existing `BAC - placeholder bugs` / `CVIE placeholder bugs` per-customer bug-parking versions. Different mechanism, different purpose — they are not per-product next-version placeholders and must not be renamed, reused or created by this skill.
```
ValentinaIancu-Gini, line 160: [Suggestion] We should mention the iOS unknown as well.

.claude/skills/gini-release/phase-1-rc.md
```diff
@@ -0,0 +1,203 @@
...
+- **Labels:** PP tickets carry `iOS` and `mobile`; HEAL is often unlabelled. Match the most recent ticket.
+- **Description** — three sections:
+  1. `**Issue Summary**` — the line "Here is the list of tickets for the release." then one `https://ginis.atlassian.net/browse/<TICKET>` link per ticket in the release. Find them with `fixVersion in ("<version-name>", …)`.
+  2. `**Listed Releases**` — one Jira release-report link per fix version: `https://ginis.atlassian.net/projects/<KEY>/versions/<id>/tab/release-report-all-issues`
+  3. `**Attachments**:` — "Build for testing can be found here:" plus the Firebase App Distribution console link and the tester-app link for the example app. **Ask the user** for both — they come from the CI/Firebase build.
+
+Also attach a **QR code of the tester-app link** as an image (LLMs can generate one on request); QA scans it directly from the ticket to install the build without typing the URL. And, if the release includes `GiniBankSDK` or `GiniHealthSDK`, link the successful XCFramework workflow run(s) from step 3 so QA has a reference build for the frameworks side.
```
ValentinaIancu-Gini, line 184: [Clarification] Where we mention the Firebase console link?

.claude/skills/gini-release/phase-1-rc.md
```diff
@@ -0,0 +1,203 @@
...
+
+**Drive status via `transitionJiraIssue`** — resolve transitions per ticket by name, case-insensitively via `getTransitionsForJiraIssue`. Ids collide across projects and can silently move a ticket to `Cancelled`. Move the ticket to `In Progress`, then to `Waiting for QA` once the build links and QR code are on it.
+
+Assign the RC ticket to the QA engineer.
```
ValentinaIancu-Gini, line 194: [Clarification]Is this needed? Who is the QA?

.claude/skills/gini-release/phase-2-tag-and-publish.md
```diff
@@ -0,0 +1,268 @@
...
+2. the version-bump PR (opened in step 4 below) has been approved by a reviewer.
+
+If the bump PR hasn't been opened yet, treat this as two sub-gates: QA sign-off before starting the bumps, then reviewer approval before pushing tags.
+
+Ask the user to confirm each; **never infer**. A tag pushed early triggers a real release.
+
+QA-failure paths:
+
+- **Showstopper:** release postponed. Add details to the RC ticket and stop.
+- **Minor:** release not postponed. Create bug tickets with the release version set as `affectedVersion` and proceed once fixed.
```
ValentinaIancu-Gini, line 42: [Clarification] Do we still need this part if we split it and it will be manually triggered by us? OR should we have something in the RC tickets that triggers this automatically?

.claude/skills/gini-release/phase-2-tag-and-publish.md
```diff
@@ -0,0 +1,268 @@
...
+
+```bash
+make lint scheme=GiniBankSDK      # or GiniCaptureSDK, GiniHealthSDK, etc.
+```
+
+`make lint` validates **compilation only**, not style. Run swiftlint separately if there are style concerns:
+
+```bash
+swiftlint --fix BankSDK/GiniBankSDK/Sources
+```
```
ValentinaIancu-Gini: [Clarification] Do we really want to run this here? I would suggest running it only when opening PRs.

.claude/skills/gini-release/phase-2-tag-and-publish.md
```diff
@@ -0,0 +1,268 @@
...
+
+Also run unit tests for the touched packages.
```
ValentinaIancu-Gini: [Clarification] I think we should skip this from here. Anyway, the tests are triggered when the push is done, next step.

.claude/skills/gini-release/phase-2-tag-and-publish.md
```diff
@@ -0,0 +1,268 @@
...
+<RC-ticket-id>
+```
+
+`<Package>` is the package name (`GiniBankSDK`). Use the RC ticket of that package's side; for `GiniUtilites` in a both-sides release include both ids. Valid commit types are in `.git-stuff/commit-msg-template.txt` — use `feat`, not `feature`.
+
+Push the release branch (no tags yet), then open a PR from it (or from an RC branch cut off it) **into the branch it came from** — usually `main`, or the version branch that carries the older major line. Follow the PR-description rules in `AGENTS.md`.
+
+**Ask the user which reviewer to assign** — do not guess. The bump is not self-reviewed; approval on this PR is one of the two gates in step 1.
+
+Wait for approval before step 5.
```
ValentinaIancu-Gini: [Clarification] How the waiting part will be executed or the next steps after manual approval?

.claude/skills/gini-release/phase-2-tag-and-publish.md
```diff
@@ -0,0 +1,268 @@
...
+If the bump PR hasn't been opened yet, treat this as two sub-gates: QA sign-off before starting the bumps, then reviewer approval before pushing tags.
+
+Ask the user to confirm each; **never infer**. A tag pushed early triggers a real release.
+
+QA-failure paths:
+
+- **Showstopper:** release postponed. Add details to the RC ticket and stop.
+- **Minor:** release not postponed. Create bug tickets with the release version set as `affectedVersion` and proceed once fixed.
+
+## 2. Bump versions on the release branch, in dependency order
```
ValentinaIancu-Gini, line 44: [Clarification] There are other places where we should bum versions. For example in BankAPILibrary documentation: 
```
We provide a sample test case [here](https://github.com/gini/gini-mobile-ios/blob/GiniBankAPILibrary%3B4.1.0/BankSDK/GiniBankSDKExample/Tests/IntegrationTests/TransferSummary/TransferSummaryIntegrationTest.swift) to verify that extraction transfer summary sending works. You may use it along with the example pdf and json files as a starting point to write your own test case.

```

DocumentServiceTests in BankAPILibrary -> ` apiLibVersion: "4.3.0",` inside `testLogErrorEvent` method.

.claude/skills/gini-release/phase-2-tag-and-publish.md
ValentinaIancu-Gini: [Clarification] How the waiting part will be executed or the next steps after manual approval?

.claude/skills/gini-release/SKILL.md
```diff
@@ -1,298 +1,33 @@
...
-  rule in `create_release_tags`, the `GiniBankSDKPinning` comment in `bank-sdk.check.yml`, and the
-  `capture-sdk-pinning-ios` repo reference are all vestigial.
-- **"Make sure you are on the `main` branch" / "Commit and push the changes to `main`"** — most releases
-  actually run from a `release/<theme>` branch and reach `main` by PR. See step 3.
-- **It does not mention `RELEASE-ORDER.md` being hand-maintained.** Unlike Android, where the equivalent
-  file is generated by a Gradle task and must never be edited manually, **iOS's `RELEASE-ORDER.md` has no
-  generator** — if a dependency changes, edit it by hand in the same commit.
-- **It says "documentation file" without saying which.** It is `Documentation/source/Installation.md`, and
-  only two packages have one. See step 4.3.
+- The [Mobile Release Process](https://ginis.atlassian.net/wiki/spaces/PLMO/pages/83689511/Mobile+Release+Process) Confluence page is the source of truth — read it if anything in the phase files seems out of date. `RELEASE-ORDER.md` follows the dependency graph in `CLAUDE.md`.
```
ValentinaIancu-Gini: [suggestion] I understood this needs to be updated in the order of steps. Can we please update it or remove it from here? We have redundancy anyway.

ValentinaIancu-Gini (commented):
@zladzeyka I finished my review on this PR and added a couple of comments. Thank you!

.claude/skills/gini-release/phase-1-rc.md
zladzeyka: Good point — you're right that Jira releases are usually created when the release is planned, not at RC time. I'll reframe step 5 as **verify the versions exist, create only if missing** (same pattern as step 6 for placeholders), and drop the "create" language from the primary flow.

.claude/skills/gini-release/phase-1-rc.md
zladzeyka: Ties into your comment on line 115 — assignment normally happens per-ticket during the sprint, not in one batch here. I'll rewrite this as a **check step**: run `fixVersion in ("<version>", …)` and if the list is empty or looks incomplete, flag it back to the user rather than fixing it silently. The skill catches "someone forgot to set fixVersion" but doesn't own the assignment itself.

.claude/skills/gini-release/phase-1-rc.md
zladzeyka: Do you mean the `iOS Unknown` fix version? I'd like to add the exact name to the "do not touch" list alongside `BAC - placeholder bugs` / `CVIE placeholder bugs` — could you confirm the name and which project(s) it lives in (PP, HEAL, both)?

.claude/skills/gini-release/phase-1-rc.md
zladzeyka: It's in step 7, line 176, inside the `**Attachments**:` bullet — "Build for testing can be found here:" + the Firebase App Distribution console link + the tester-app link, both marked **Ask the user**. Line 178 only adds the extras that ride on top (QR code + XCFramework run link). I can pull the Firebase vs tester-app distinction out of the bullet into its own line if it's getting lost — want me to?

.claude/skills/gini-release/phase-1-rc.md
zladzeyka: The assignment step is what actually gets the ticket onto the QA engineer's board — moving to `Waiting for QA` alone doesn't reassign it. The skill can't know who's on QA duty for a given sprint, so it asks the user. If you'd rather we default to a fixed person (or leave unassigned and let QA pick it up), let me know and I'll swap it in. I'd lean on asking, since it also covers vacation coverage.

.claude/skills/gini-release/phase-2-tag-and-publish.md
zladzeyka: I'd keep the manual gate for now. Phase 2 pushes irreversible tags, and someone can invoke `/gini-release` without having done phase 1 in the same session (Claude has no memory across sessions and the RC ticket status isn't polled). A two-line confirmation ("QA signed off? bump PR approved?") is cheap and prevents an accidental early tag push. Auto-triggering from the RC ticket status would be nicer but needs Jira → Claude wiring we don't have today — happy to do it as a follow-up if there's appetite.

.claude/skills/gini-release/phase-2-tag-and-publish.md
zladzeyka: Great catch, thank you. I'll add a fifth bump target for **doc URLs with tag paths and hardcoded version strings in tests**, with a grep-first pass so the skill scans each touched package for the *previous* version before committing. Adding both examples you flagged:

- `BankAPILibrary/…` README/doc links like `.../GiniBankAPILibrary%3B4.1.0/…`
- `DocumentServiceTests.testLogErrorEvent` → `apiLibVersion: "4.3.0"`

If there are other lurking spots you know of, drop them here — otherwise I'll rely on the grep to flag them.

.claude/skills/gini-release/phase-2-tag-and-publish.md
zladzeyka: Agree — `swiftlint --fix` doesn't belong in the release path, PRs cover it. Removing.

.claude/skills/gini-release/phase-2-tag-and-publish.md
zladzeyka: Agree, the bump PR runs them on push. Removing.

.claude/skills/gini-release/phase-2-tag-and-publish.md
zladzeyka: The "wait" is not a poll — the skill stops here, reports what it did (open PR link, what's still pending) and exits. The user resumes phase 2 in a new invocation once the bump PR is approved (`/gini-release <RC-ticket>` again, or `/gini-release continue`). I'll make that explicit in the wording so it's clear nothing runs in the background.

.claude/skills/gini-release/SKILL.md
zladzeyka: Two readings — happy to do either or both, want to confirm before I touch it:

(a) Drop this bullet from `SKILL.md`. The phase files are self-contained and the Confluence link is already inside the RC and tag-and-publish files at the top.
(b) Keep the bullet but caveat it as "Confluence page is being updated to match the two-phase split".

Which do you prefer? Or is there a third thing I'm missing?

MozhganPeivandianSharbaf:
Hi @zladzeyka  @ValentinaIancu-Gini I have a question,  in my point of view the phase 1 is rc and second phase is bumping and tags since publish will happen after reviewing I preferred to have phase 2 with this name [phase-2-bump-and-tag](https://github.com/gini/gini-mobile-android/pull/984/changes) could you let me know why you changed the name? as I can see you are doing the same in second phase

.claude/skills/gini-release/phase-2-tag-and-publish.md
zladzeyka: Follow-up — grepped the repo before pushing and found my paths were off. Corrected in `226718ea`:

- `BankAPILibrary/GiniBankAPILibrary/README.md` doesn't have those links — the real file is `Documentation/source/Getting started.md` (line 317, `GiniBankAPILibrary%3B4.1.0/…`, stale by two patches).
- Missed `BankSDK/GiniBankSDK/README.md` entirely. That one has drifted **three majors** — still points at `GiniBankSDK;3.3.0` while current bank is 4.3.0. Added it to the known-spots list with a note that it drifts historically.
- Also flipped the grep hint from "previous version" to "current version, before bumping the version file" — tag-path URLs embed the version *being released*, not the previous one.

Nothing to redo on your side — just wanted to note the fix landed.

.claude/skills/gini-release/SKILL.md
zladzeyka: Dropped in `13580759`. The Confluence link is still at the top of each phase file, so no info is lost.

.claude/skills/gini-release/phase-2-tag-and-publish.md
zladzeyka: Same question as [the thread on line 105 above](https://github.com/gini/gini-mobile-ios/pull/1245#discussion_r3955018028) — copying the answer here so it doesn't get lost:

There is no background waiting or polling. Phase 2 reaches this line and **stops**: it reports the open bump PR URL, what's still pending (bump PR approval, QA sign-off if either is still open), and ends the session. Claude has no memory across sessions, so it can't sit and watch the PR.

The user resumes phase 2 by re-invoking the skill — `/gini-release <RC-ticket>` — once the bump PR is approved. On resume, the skill re-confirms both gates from step 1 (QA signed off? bump PR approved?) with the user before continuing to step 5 (`create_release_tags`). If either gate is still open on the resume, it stops again.

I made this explicit in `bc9aef51d` — the "wait for approval before step 5" line is now:

> **Then stop and exit.** There is no background waiting or polling — report the open PR link plus what's still pending (bump PR approval, QA sign-off if still open), and end the session. The user resumes phase 2 by re-invoking the skill (`/gini-release <RC-ticket>`) once the PR is approved; on resume, re-confirm both gates in step 1 before continuing to step 5.

zladzeyka:
Good catch — the divergence is intentional. iOS phase 2 covers more than just bump and tag: the "publish" half is meaningful work that Android delegates to Gradle/Sonatype but iOS has to script by hand. Roughly 9 of the 14 sections in `phase-2-tag-and-publish.md` are the publish chain — XCFramework builds via `bank-sdk.build.xcframeworks` / `health-sdk.build.xcframeworks`, draft GitHub releases in mono + individual repos, a hard user-review gate before publishing, CocoaPods podspec push for GiniBankSDK, Jira release publication, RC close-out, merge back to `main`, and the `#mobile-releases` announcement. Naming it `bump-and-tag` on iOS would undersell it and likely send a reader looking for a `phase-3` that doesn't exist.

Two alternatives if you'd prefer closer alignment:

1. Rename to `phase-2-bump-tag-and-publish.md` — one file, more explicit label, trivial change.
2. Split iOS into `phase-2-bump-and-tag.md` + `phase-3-publish.md` for full structural parity with Android's naming — bigger refactor, would need a `/gini-release publish` invocation on top of the current two.

Happy to do either — mildly prefer keeping the current name since the whole tag → publish chain gates the same session state (RC ticket, XCFramework runs, draft URLs the user reviews) and splitting would require re-entry with `--continue` semantics. What do you think?

qnaveed87 (approved):
@zladzeyka thank you for the great work 

Small gap noticed while releasing phase 2: step 12 mentions the TestFlight link, but no earlier step uploads to TestFlight. Does adding a "dispatch TestFlight workflow" step make sense here ?

MozhganPeivandianSharbaf:
@zladzeyka  lets have this naming as you have it right now, Thank you

.claude/skills/gini-release/phase-1-rc.md
ValentinaIancu-Gini: Yes that version.

zladzeyka: Ok I will adapt it

zladzeyka:
@qnaveed87 addressed your TestFlight-gap catch in [`1cce12ac1`](https://github.com/gini/gini-mobile-ios/pull/1245/commits/1cce12ac1) — added a new phase-2 step 12 that manually dispatches both [`bank-sdk.publish.example.app.testflight.yml`](https://github.com/gini/gini-mobile-ios/actions/workflows/bank-sdk.publish.example.app.testflight.yml) and [`health-sdk.publish.example.app.testflight.yml`](https://github.com/gini/gini-mobile-ios/actions/workflows/health-sdk.publish.example.app.testflight.yml) before the Slack announcement, so the build being announced actually contains what just shipped instead of the last quarterly cron run. Skips either workflow if that SDK isn't in the release. Renumbered the two later steps (merge+Slack → 13, report → 14) and updated the Slack step to explicitly reference the freshly-triggered build.

PR #1247, opened 2026-08-25
Create gini test skill

MozhganPeivandianSharbaf:
## Pull Request Description

Moves the Xray test-case tooling into a single `gini-test-skill` folder and adds a script
that exports an Xray Test Set to CSV — the direction Xray Cloud itself does not offer.

N/A — no Jira ticket; tooling change only.

**Why the exporter is needed.** Xray Cloud has no CSV export that carries step text. A
test case there is two pieces: the Jira issue (key, summary, status, labels), which Jira
owns, and the steps (Step / Data / Expected Result), which Xray stores outside Jira. So
Jira's `Export → CSV — all fields` returns hundreds of Jira columns and none of the steps,
the Tests panel on a Test Set offers no export at all, and the Xray Test Case Importer only
runs inbound. The API is the only route to the steps.

**What changed**

- `.claude/skills/generate-xray-tests/` → `.claude/skills/gini-test-skill/`. The generator
  is now `generate-xray-tests.md` and its content is **unchanged** — git records it as a
  rename, so its history is preserved. `references/example-tests.csv` moved with it,
  also unchanged.
- New `.claude/skills/gini-test-skill/xray_export_testset_csv.py`. It authenticates against
  Xray Cloud (`POST /api/v2/authenticate`), queries `getTests` over GraphQL scoped by
  `issue in testSetTests("<KEY>")`, pages through the results, then writes one CSV row per
  test step.

**How the exporter works.** Credentials come from `~/.config/gini/xray.env`, never from the
command line, so they cannot land in shell history. GraphQL rather than REST because the
REST `/steps` endpoint omits preconditions entirely. Output columns are
`Issue Id, Summary, Test Type, Step, Data, Expected Result, Jira Key, Total Steps` — the
first six are the Xray Test Case Importer layout in its original order, so an existing
importer column-mapping still works; `Jira Key` repeats on every row so the file stays
sortable in a spreadsheet.

**Trade-offs taken**

- **No `Preconditions` column.** Xray's importer expects that column to hold an existing
  Precondition issue *key*; free text there fails the import with "Precondition type and
  test type mismatch". The script reads preconditions anyway and warns, naming each case
  that has one, so nothing is dropped silently.
- **Values are written verbatim, not sanitised for spreadsheets.** A field starting with
  `=`, `+`, `@` or `-` is a formula to Excel. Escaping it would break re-import into Xray,
  and expected-result text legitimately starts with `-` (dash-bulleted observations). This
  is documented in the module docstring; sanitise downstream if you need a spreadsheet-safe
  copy.
- **The accessibility split is a text match** on the summary (accessibility / VoiceOver /
  TalkBack) or an `accessibility` label — not an Xray field. Good enough in practice, but
  worth eyeballing.

## Notes for Reviewers

No app code is touched — this PR only adds files under `.claude/skills/`, so there is
nothing to build and no simulator run applies.

**How it was verified**

- Ran the exporter against a real Test Set end to end: authentication succeeded, all cases
  the Test Set reports were fetched (`Fetched: N of N`), and both CSVs were written with the
  expected case and step-row counts.
- Parsed both output files with Python's `csv` module — not grep — to confirm every data row
  carries a `Jira Key`, that the distinct key count matches the case count, and that no
  accessibility case leaked into the functional file.
- Cross-checked the scope independently in the Jira issue navigator with
  `issue in testSetTests("<KEY>")`; the count matched the API's.
- Confirmed the generator move is a pure rename by checksumming it against its previous
  content on `main`.

**Try it yourself**

```bash
python3 .claude/skills/gini-test-skill/xray_export_testset_csv.py \
  --test-set <TEST-SET-KEY> --prefix <feature> --out-dir ./out
```

Needs `XRAY_CLIENT_ID` / `XRAY_CLIENT_SECRET` in `~/.config/gini/xray.env` (mode `600`,
outside any repo). Our instance is EU, so the base URL is
`https://eu.xray.cloud.getxray.app`. Setup is documented on Confluence:
[Xray test cases export](https://ginis.atlassian.net/wiki/spaces/PLMO/pages/1885470721/Xray+test+cases+export).

**Known limitations / follow-up**

- No unit tests. The script is a thin wrapper over the Xray API and its only meaningful
  paths are network calls; it was verified by running it and parsing its output instead.
- **It does not fetch Test Repository folder paths.** If you need to check that a Test Set
  has not drifted into an archive folder, run a separate `getTests { folder { path } }`
  query. Worth adding to the script later.
- `generate-xray-tests.md` still carries `name: generate-xray-tests` in its frontmatter but
  is no longer at a `SKILL.md` path, so it does not register as an invocable skill. Same is
  true of the folder as a whole — it is a tooling folder, not a skill, until a `SKILL.md`
  is added.
- The Android side of this is not in this PR.


.claude/skills/gini-ui-test-work-flow-skill/xray_export_testset_csv.py
```diff
@@ -0,0 +1,299 @@
...
+    else:
+        groups = [
+            ("", [t for t in tests if not is_accessibility(t)]),
+            ("-accessibility", [t for t in tests if is_accessibility(t)]),
+        ]
+
+    for suffix, group in groups:
+        if not group:
+            continue
+        path = os.path.join(args.out_dir, "%s%s.csv" % (args.prefix, suffix))
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: Intentional — writing an empty CSV would just be noise. I fixed the contract instead: the docstring now says it writes up to two files, and the script prints the path of every file it actually wrote, so use that rather than assuming both exist.

.claude/skills/gini-ui-test-work-flow-skill/xray_export_testset_csv.py
```diff
@@ -0,0 +1,299 @@
...
+                rows_written += 1
+                continue
+
+            for step_index, step in enumerate(steps):
+                first = step_index == 0
+                writer.writerow([
+                    tc_id,
+                    summary if first else "",
+                    test_type,
+                    (step.get("action") or "").strip(),
```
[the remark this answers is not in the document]
MozhganPeivandianSharbaf: Not escaping this on purpose. The first six columns have to stay re-importable into Xray, and expected results legitimately start with `-` (dash-bulleted observations), so prefixing would corrupt real data. The input is our own Xray content, not untrusted. I documented the behaviour in the docstring — if we need a spreadsheet-safe copy we can sanitise downstream.

ValentinaIancu-Gini:
@MozhganPeivandianSharbaf I would like to ask if you renamed/moved the skill I created for the X-ray test generation.  I'm asking this because this will impact Yuliia as well since shw was using from this repository.

Also, can you please check the comments from Copilot and resolve them?

.claude/skills/gini-ui-test-work-flow-skill/xray_export_testset_csv.py
```diff
@@ -0,0 +1,314 @@
...
+Step and result text is written EXACTLY as Xray holds it, with no escaping beyond
+normal CSV quoting. A field that begins with "=", "+", "@" or "-" is therefore
+treated as a formula by Excel and Google Sheets. That is deliberate: prefixing
+such fields to defuse them would corrupt the file for re-import into Xray, and
+step text legitimately starts with "-" (dash-bulleted expected results). Treat
+the output as data exported from Xray rather than as a trusted spreadsheet, and
+if you need a spreadsheet-safe copy, sanitise it downstream instead of here.
+
+Credentials are read from a file (never passed on the command line, so they do
+not land in shell history). Default location: ~/.config/gini/xray.env
```
zladzeyka, line 45: [Clarification]: Let's imagine I use, will I be asked to provide the creds in `~/.config/gini/xray.env` or is it a precondition for a skill

zladzeyka (commented):
Thank you @MozhganPeivandianSharbaf 🙏
looks good just a tiny clarification

.claude/skills/gini-ui-test-work-flow-skill/xray_export_testset_csv.py
MozhganPeivandianSharbaf: It's a precondition — set it up once, the skill never asks you for it.

Create `~/.config/gini/xray.env` with:

```
XRAY_CLIENT_ID=...
XRAY_CLIENT_SECRET=...
```

You get those two values in Xray global settings → API Keys. Steps are here: [Xray test cases export](https://ginis.atlassian.net/wiki/spaces/PLMO/pages/1885470721/Xray+test+cases+export).

If the file is missing, the script stops right away and prints what to create.

zladzeyka: thank u!

zladzeyka (approved):

.claude/skills/gini-ui-test-work-flow-skill/xray_export_testset_csv.py
```diff
@@ -0,0 +1,314 @@
...
+                ])
+                rows_written += 1
+    return rows_written
+
+
+def main():
+    parser = argparse.ArgumentParser(description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument("--creds", default=DEFAULT_CREDS, help="credentials file (default: %s)" % DEFAULT_CREDS)
+    parser.add_argument("--out-dir", default=".", help="directory to write the CSVs into")
+    parser.add_argument("--prefix", default="credit-note", help="output filename prefix")
```
qnaveed87, line 255: [Request] Can you please use a generic name here instead of `credit-note`

.claude/skills/gini-ui-test-work-flow-skill/xray_export_testset_csv.py
```diff
@@ -0,0 +1,314 @@
+#!/usr/bin/env python3
+"""
+Export Xray Cloud manual test cases to CSV, one row per test STEP.
+
+Writes UP TO two files:
+
+    <prefix>.csv                 functional tests
+    <prefix>-accessibility.csv   accessibility tests
+
+Splitting into the two is the default. A group with no matching tests is skipped
```
qnaveed87, line 10: [Suggestion]: Can you please update this  `Splitting into the two is the default`  it looks incomplete

qnaveed87 (changes requested):
@MozhganPeivandianSharbaf Thank you for the good work 

I left 2 small comments

PR #1253, opened 2026-09-03
fix: Restore GiniCaptureResultsDelegate source compatibility (4.5.0 re-release)

zladzeyka:
## Pull Request Description

Re-ship of `GiniBankSDK` and `GiniCaptureSDK` **4.5.0** after the original 4.5.0 was withdrawn earlier today. The original release shipped a source-compatibility break: `GiniCaptureResultsDelegate` gained a required method `giniCaptureDidRequestSchedulePayment(result:)` with no default implementation, so every host-app type conforming to the protocol failed to compile against 4.5.0 until it added the method. Discovered by Valentina while upgrading the demo app.

[PP-3497](https://ginis.atlassian.net/browse/PP-3497)

### The fix

Mark the new method `@objc optional` on the existing `@objc` `GiniCaptureResultsDelegate` protocol (idiomatic UIKit-style delegate evolution) and switch the one SDK-internal call site to optional-method dispatch. Integrators upgrading 4.4.x → 4.5.0 keep compiling without changes. Integrators who want to handle Schedule Payment implement the method as they would have under the original 4.5.0.

Ships as **`GiniBankSDK` 4.5.0 + `GiniCaptureSDK` 4.5.0** — same version number as the withdrawn release. `GiniBankAPILibrary` 4.5.0 and `GiniUtilites` 2.6.0 are unaffected and were not withdrawn.

### Decision trail (spec + commits)

The branch carries a `docs(specs)` commit with the full plan (`specs/PP-3497-feature.md`), an Option A implementation commit (Swift `public extension` default impl + `@objc` drop), and a swap to Option C (`@objc optional`). The intermediate Option A commit is preserved so the review context is visible; feel free to squash-merge — the end state is what ships.

### Integration matrix — what integrators experience on 4.5.0

| `paymentScheduleHintEnabled` | `giniCaptureDidRequestSchedulePayment(result:)` implemented? | What happens |
|---|---|---|
| `false` (default) | — | *Schedule Payment* state on the payment-hint bottom sheet never appears. Callback never fires. **Zero impact.** |
| `true` | Yes | Sheet appears → user picks *Schedule Payment* → callback fires → integrator handles it. **Works.** |
| `true` | No | Sheet appears → user picks *Schedule Payment* → optional dispatch silently no-ops → user is stuck. **UX dead-end** — integrator responsibility to implement the callback once they turn the flag on. |

### Integration on the integrator's side

**Existing 4.4.x conformer — nothing to do to keep compiling on 4.5.0:**

```swift
class ScreenAPICoordinator: NSObject, GiniCaptureResultsDelegate {
    func giniCaptureAnalysisDidFinishWith(result: AnalysisResult) { … }
    func giniCaptureDidCancelAnalysis()                            { … }
    func giniCaptureDidEnterManually()                             { … }
}
```

**Adopting Schedule Payment on 4.5.0:**

```swift
class ScreenAPICoordinator: NSObject, GiniCaptureResultsDelegate {
    func giniCaptureAnalysisDidFinishWith(result: AnalysisResult) { … }
    func giniCaptureDidCancelAnalysis()                            { … }
    func giniCaptureDidEnterManually()                             { … }

    // New — signature must match exactly; @objc optional makes wrong labels a silent miss.
    func giniCaptureDidRequestSchedulePayment(result: AnalysisResult) {
        dismissSDK { [weak self] in
            self?.presentScheduledTransfer(extractions: result.extractions)
        }
    }
}
```

Then set `GiniBankConfiguration.shared.paymentScheduleHintEnabled = true`.

## Notes for Reviewers

- Documentation-only migration required for existing integrators, and even that only if they want to adopt Schedule Payment. Anyone leaving the flag at its default (`false`) sees zero behavior change.
- The one silent-signature-drift trap of `@objc optional`: if an integrator writes a slightly-wrong label (`_ result:` instead of `result:`), the method compiles as an unrelated method and the SDK never calls it. Called out explicitly in the release note snippet in the spec.

### Testing

- **New tests** — `CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/GiniCaptureResultsDelegateDefaultImplTests.swift` — three focused tests:
  - Compile-guarded conformance (a private `Pre45CaptureResultsDelegate` implements only the three pre-4.5 methods and instantiates as `GiniCaptureResultsDelegate` — if the fourth method loses `@objc optional`, this file stops building).
  - Optional dispatch resolves to `nil` at the call site when the delegate does not implement the method.
  - Optional dispatch invokes the implementation when the delegate does implement it.

- **Local runs** (iPhone 17 Pro / iOS 26.5, closest local match to CI's iOS 26.2):
  - `GiniCaptureSDKTests/GiniCaptureResultsDelegateDefaultImplTests` → **3/3 pass** (0.008 s).
  - `GiniBankSDKTests/NetworkingScreenApiCoordinatorTests` → **95/95 pass** (10.5 s) — proves the SDK-side optional dispatch still invokes an implementing delegate correctly, so R3 in the spec is covered.

- **CI parity**: please re-run on `iPhone 16 / iOS 26.2` per repo standards; couldn't run locally as 26.2 runtime isn't installed here.

### Not addressed here

- **Confluence migration page for 4.4 → 4.5** — HTML staged at `specs/migrate-4.4-to-4.5-draft.html`, still blocked on Atlassian connector re-consent for Confluence write. To be posted separately.
- **XCFramework archive + CocoaPods podspec** — produced fresh for the re-shipped 4.5.0 through the normal release pipeline once this PR merges.

[PP-3497]: https://ginis.atlassian.net/browse/PP-3497

---

Supersedes #1252 (auto-closed by GitHub when the head branch `release/GiniBankSDK-4.5.1` was renamed to `release/GiniBankSDK-4.5.0_fix`; GitHub does not allow reopening a PR whose head ref no longer exists). Same commits, same content.

### DIFF
--- a/BankSDK/GiniBankSDK/Package-release.swift
+++ b/BankSDK/GiniBankSDK/Package-release.swift
@@ -16,7 +16,7 @@ let package = Package(
     dependencies: [
         // Dependencies declare other packages that this package depends on.
         // .package(url: /* package url */, from: "1.0.0"),
-        .package(name: "GiniCaptureSDK", url: "https://github.com/gini/capture-sdk-ios.git", .exact("4.5.0")),
+        .package(name: "GiniCaptureSDK", url: "https://github.com/gini/capture-sdk-ios.git", .exact("4.5.1")),
     ],
     targets: [
         // Targets are the basic building blocks of a package. A target can define a module or a test suite.
--- a/BankSDK/GiniBankSDK/Package-release.swift
+++ b/BankSDK/GiniBankSDK/Package-release.swift
@@ -16,7 +16,7 @@ let package = Package(
     dependencies: [
         // Dependencies declare other packages that this package depends on.
         // .package(url: /* package url */, from: "1.0.0"),
-        .package(name: "GiniCaptureSDK", url: "https://github.com/gini/capture-sdk-ios.git", .exact("4.5.1")),
+        .package(name: "GiniCaptureSDK", url: "https://github.com/gini/capture-sdk-ios.git", .exact("4.5.0")),
     ],
     targets: [
         // Targets are the basic building blocks of a package. A target can define a module or a test suite.
--- a/BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
+++ b/BankSDK/GiniBankSDK/Sources/GiniBankSDK/Core/GiniBankNetworkingScreenApiCoordinator.swift
@@ -852,7 +852,7 @@ internal extension GiniBankNetworkingScreenApiCoordinator {
                                             document: documentService.document,
                                             candidates: extractionResult.candidates)
         sendAnalyticsEventSDKClose()
-        resultsDelegate?.giniCaptureDidRequestSchedulePayment(result: analysisResult)
+        resultsDelegate?.giniCaptureDidRequestSchedulePayment?(result: analysisResult)
         documentService.resetToInitialState()
     }
 
--- a/BankSDK/GiniBankSDK/Sources/GiniBankSDK/GiniBankSDKVersion.swift
+++ b/BankSDK/GiniBankSDK/Sources/GiniBankSDK/GiniBankSDKVersion.swift
@@ -5,4 +5,4 @@
 //  Copyright © 2024 Gini GmbH. All rights reserved.
 //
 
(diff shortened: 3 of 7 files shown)


ValentinaIancu-Gini (approved):
@zladzeyka looks good

PR #1262, opened 2026-09-07
feat: Remove bottom navigation bar adapter API and internal wiring

zladzeyka:
## Pull Request Description

Removes the dead bottom-navigation-bar customization surface across `GiniCaptureSDK` and `GiniBankSDK`. The feature has been dark-launched-off in 4.5.x — every enable knob on `GiniConfiguration` / `GiniBankConfiguration` was already `internal let` and hard-coded to `false` / `nil`, so no integrator can wire an adapter at runtime today. This PR completes the cleanup by deleting the still-`public` protocol surface, the two `public` button enums those protocols exposed, the internal default implementations and views, the coordinator/VC branches that were unreachable behind `if bottomNavigationBarEnabled`, and the example-app demonstrations of an integration path that no longer works.

[PP-2273]

**How it was done** — followed `specs/PP-2273-feature.md`, deletion in dependency order so each commit compiled: (1) example-app custom adapters, (2) de-branch consumer VCs, (3) delete configuration properties, (4) delete default implementations and internal `UIView` subclasses, (5) delete public protocols, (6) doc/resource/pbxproj sweep. `ReviewViewController` collapsed from four constraint arrays to two (dropped the `WithBottomBar` variants and simplified the height-multiplier helpers); `OnboardingViewController` lost the `skipBottomBarButton` outlet and its layout branch. Mirror-walk regression tests added in each SDK to catch accidental re-introduction. **Version bump deferred** — this PR ships the code deletion only; `GiniCaptureSDKVersion` / `GiniBankSDKVersion` stay at `4.5.0` and the customer-facing 5.0.0 migration note moves to a follow-up ticket that cuts the 5.0.0 release branch.

## Notes for Reviewers

</br>

**Compile:** `xcodebuild build -workspace GiniMobile.xcworkspace -scheme GiniCaptureSDK` and same for `GiniBankSDK` — both `** BUILD SUCCEEDED **` on iPhone 17 Pro simulator (Xcode 26.5 SDK).

**Tests:** `xcodebuild test` against both SDK schemes:
- `GiniCaptureSDKTests/ReviewViewControllerTests` (6 tests) — pass, including the collapsed constraint-multiplier assertions.
- `GiniCaptureSDKTests/GiniConfigurationBottomNavRegressionTests` (1 test, new) — Mirror walk over `GiniConfiguration()` asserting no stored property name contains `bottomNavigation` or ends in `NavigationBarBottomAdapter`.
- `GiniBankSDKTests/GiniBankConfigurationBottomNavRegressionTests` (1 test, new) — same for `GiniBankConfiguration`.
- `GiniBankSDKExampleTests` (`SettingsViewModelTests` etc.) — blocked locally by a broken SwiftLint / SourceKit environment (`sourcekitdInProc.framework` load fails, `swiftlint` crashes with SIGILL). Same issue previously noted; will pass in CI where SwiftLint is healthy.

**Mechanical acceptance:** `grep -rln 'bottomNavigation\|BottomNavigation\|bottomNav' CaptureSDK/GiniCaptureSDK/Sources/ BankSDK/GiniBankSDK/Sources/` returns **0** hits. Same grep across `Tests/` returns only the two new regression files (where the strings appear inside `#expect(...)` arguments).

**Simulator eyeball — needed from a reviewer:** the `ReviewViewController` constraint collapse and `OnboardingViewController` skip-button removal have not been visually verified in a simulator yet. Recommend a manual pass through Onboarding (all pages), Review (portrait + landscape, iPhone SE 3 + iPhone 17 Pro Max, iPad), Help menu/tips/formats/import, Error, NoResult, Camera. This PR is opened as **draft** for that reason.

**BrowserStack:** smoke suite (`bs_run_smoke_tests.sh`) has not been run against this branch. Recommend a run on the standard two-device matrix (`iPhone 17-26`, `iPhone 16-18`) before merge.

**Public API impact — source-breaking:** fourteen `public` types are deleted — twelve `public protocol`s and two `public enum`s:

- **CaptureSDK protocols (6):** `CameraBottomNavigationBarAdapter`, `ImagePickerBottomNavigationBarAdapter`, `ReviewScreenBottomNavigationBarAdapter`, `ErrorNavigationBarBottomAdapter`, `OnboardingNavigationBarBottomAdapter`, `HelpBottomNavigationBarAdapter`.
- **BankSDK protocols (6):** `DigitalInvoiceOnboardingNavigationBarBottomAdapter`, `DigitalInvoiceHelpNavigationBarBottomAdapter`, `DigitalInvoiceNavigationBarBottomAdapter`, `DigitalInvoiceSkontoNavigationBarBottomAdapter`, `SkontoNavigationBarBottomAdapter`, `SkontoHelpNavigationBarBottomAdapter`.
- **CaptureSDK button enums (2):** `CameraNavigationBarBottomButton`, `OnboardingNavigationBarBottomButton` — used only by the deleted default implementations, but part of the public surface.

Runtime behavior for any integrator that still references one of these types is unchanged (the SDK stopped consulting adapters when `bottomNavigationBarEnabled` was moved to `internal let false`); the delta is a clean compile error at update. Migration: delete the adapter class, any assignment to the (already-`internal`) matching configuration property, and any reference to the two button enums.

**Follow-up:** version bump `4.5.0 → 5.0.0` for both SDKs, `Package-release.swift` bump, and the customer-facing migration note in `RELEASE.md` land in a separate ticket that cuts the 5.0.0 release branch.

[PP-2273]: https://ginis.atlassian.net/browse/PP-2273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CaptureSDK/GiniCaptureSDK/Tests/GiniCaptureSDKTests/GiniScreenAPICoordinator+CameraTests.swift
```diff
@@ -0,0 +1,534 @@
...
+    }
+
+    // MARK: - didCaptureAndValidate
+
+    func testDidCaptureAndValidate_callsVisionDelegateDidCapture() {
+        let document = GiniCaptureTestsHelper.loadImageDocument(named: "invoice")
+
+        coordinator.didCaptureAndValidate(document)
+
+        XCTAssertEqual(spyDelegate.didCaptureCallCount, 1)
```
qnaveed87: [Suggestion ] As we did with `GiniConfigurationBottomNavRegressionTests` with new swift Testing  can we also use it here

qnaveed87 (changes requested):
Thank you @zladzeyka for the nice work 

I left one small suggestion

qnaveed87 (approved):
@zladzeyka Looks good to me

PR #1269, opened 2026-09-15
fix(GiniInternalPaymentSDK): preserve page order in PaymentReview preview fetch

zladzeyka:
## Pull Request Description

Fixes a regression introduced by the PaymentReview SwiftUI migration in `GiniHealthSDK;6.0.0` where a multi-page PDF's preview renders pages in an arbitrary order and re-races on every screen re-open. The pre-migration `5.5.2` implementation serialised page fetches with `DispatchSemaphore` and preserved page order; the migration parallelised the work with `withTaskGroup` but dropped the ordering invariant, because `reduce(into:)` appends results in completion order and the launching `page` index is discarded on the child-task return type.

[HEAL-558](https://ginis.atlassian.net/browse/HEAL-558)

**How:** kept concurrency, re-imposed order — each child task now returns `(page - 1, cellViewModel)` and the results are written into a pre-sized `[PageCollectionCellViewModel?]` buffer by index; `compactMap` drops nil (failed-fetch) slots while preserving relative order before the batch assignment to `cellViewModels`. Body-only change in `PaymentReviewModel.fetchImages()`; no public API impact, no downstream Health SDK code change (`PaymentReviewObservableModel.fetchImages()` delegates to the fixed model). Alternative shape considered was a tuple-collect-then-sort — the pre-sized buffer is more explicit about the invariant.

## Notes for Reviewers

**Regression test added:** `PaymentReviewModelTests.fetchImagesPreservesPageOrder()` stubs `MockPaymentReviewDelegate.preview(for:pageNumber:completion:)` with inverted per-page delays (page 1 slowest, page N fastest — completion order 4, 3, 2, 1) and asserts `cellViewModels[i].preview` corresponds to page `i + 1`. Fails against the un-fixed body with the observed inverted order; passes with the fix.

**Mock change:** extended `MockPaymentReviewDelegate` with `previewDataByPage: [Int: Data]` and `previewDelaysByPage: [Int: TimeInterval]`. Zero-defaults preserve existing behavior — no other test suite needed changes and the pre-existing `"fetchImages sets isImagesLoading and calls onPreviewImagesFetched"` test still passes.

**Simulator / iOS:** iPhone 17 Pro / iOS 26.5 (iPhone 16 / iOS 26.2 not installed locally).

**Manual checks (Health SDK example):**
- Payment Review with a mixed-readability PDF (one Gini-readable page + N raster-scan pages) — pages now render in original PDF order.
- Dismiss and re-open — order remains stable across re-opens.

**Automated:**
- `GiniInternalPaymentSDK` full test target — 258 tests / 33 suites passed.
- `GiniHealthSDK` scheme — BUILD SUCCEEDED.

**Follow-ups (deliberately not in scope):**
- `fetchImages()` still traps at runtime if `document.pageCount == 0` (`1...0` is an invalid closed range). Pre-existing behavior in this call site.
- `cellViewModels.append(contentsOf:)` accumulates on repeat calls; current callers create a fresh model per screen open, so a second call is not observed today.

**Diagnosis document:** `specs/HEAL-558-bug.md` — full root-cause analysis produced by `/gini-fix`, included in this PR for reviewer context.

[HEAL-558]: https://ginis.atlassian.net/browse/HEAL-558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


GiniComponents/InternalPaymentSDK/GiniInternalPaymentSDK/Tests/GiniInternalPaymentSDKTests/PaymentTestMocks.swift
```diff
@@ -47,7 +49,15 @@ final class MockPaymentReviewDelegate: PaymentReviewProtocol {
...
     }
     func preview(for documentId: String, pageNumber: Int, completion: @escaping (Result<Data, GiniError>) -> Void) {
-        completion(previewResult)
+        let payload: Result<Data, GiniError> = previewDataByPage[pageNumber].map { .success($0) } ?? previewResult
+        guard let delay = previewDelaysByPage[pageNumber], delay > 0 else {
+            completion(payload)
+            return
+        }
+        Task {
+            try? await Task.sleep(nanoseconds: UInt64(delay * 1_000_000_000))
```
qnaveed87: [Suggestion] Can you please replace it with 
`Task.sleep(for: .seconds(delay))`

qnaveed87 (changes requested):
Thank you for the PR 

I have left one minor suggestion

qnaveed87 (approved):
Looks good to me

